PR #65, opened 2013-09-17
Top navigation items span over whole height

m-kuhn:


rduivenvoorde:
reworked to f37a1c8841ed839591e09fd81b18fca5432dd198

PR #71, opened 2013-09-19
Update QGIS core development text

m-kuhn:


pcav:
Dev list is duplicated.Better extract automatically from GH, or point to its relevant page; nobody is going to keep it (them) updated anyway

PR #109, opened 2013-10-04
Show windows download links as block

m-kuhn:
I'm not sure if I edited the correct file, because `make html` did not make the qgis-style.css update on my machine... Please review shortly.

### DIFF
--- a/themes/qgis-theme/static/qgis-style.less
+++ b/themes/qgis-theme/static/qgis-style.less
@@ -637,6 +637,7 @@ section.content h4 {
 }
 #downloads .accordion-inner a {
   text-decoration: none;
+  display: block;
 }
 #downloads .accordion-inner .icon-large {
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
hi matthias, though I'm fine with merging this, it is not clear what exactly this fixes?
Can you tell us plz?

m-kuhn:
Looking at the windows section of the download page ( https://qgis.org/en/site/forusers/download.html ) right now, it renders here the two OSGeo4W installer links left and right of each other, while the other two downloads are below each other. Apart from this, the square-and-arrow-link icon to the right of the links is mispositioned for the 32bit standalone item.

rduivenvoorde:
fix taken into: https://github.com/qgis/QGIS-Website/commit/d5c9e7106538feb67f629ec8ce393ba22b86b38f

PR #112, opened 2013-10-05
new qgis-gallery.png

anitagraser:
update with newer and fewer maps which show a nice variety of styles


rduivenvoorde:
hi anita,

better picture indeed!

should we not also do another image for the QGIS Users (blue) map? 
I think it is a classical example of 'red dot fever' :-)

anitagraser:
Yes that one should be replaced as well. Either by a better map or by a
nice workshop or qgis in action picture.  I will see what I can do.

PR #118, opened 2013-11-15
Update RHEL/CentOS installation instructions

m-kuhn:


rduivenvoorde:
hi matthias,
I'll pull, but please keep the dots in a header exactly the size of the header (otherwise we end up with sphinx warnings).
And I will change 'QGIS 2.0' to 'QGIS stable' (as that is what we agreed upon)

Is there no way to pull QGIS testing into RHEL of Centos?

PR #131, opened 2014-01-16
Update alldownloads.po

cdavila:


rduivenvoorde:
hi @cdavila 

sorry, but the translation of website etc takes place via transifex. The .po files are here only as backup.

see https://www.transifex.com/organization/qgis

merging this and then merging into transifex appears to be too complex.

sorry again

PR #133, opened 2014-01-24
added bugreports how2 from wiki to development index page

anitagraser:


anitagraser:
Hi @rduivenvoorde! Could you have a look at this addition?

rduivenvoorde:
hi @anitagraser 
just pulled and will build it as is.
 I think we should split up this part (development) into more pages. I will have a look into this.
What is the idea about the strings with the @ signs in it?

anitagraser:
@rduivenvoorde Good catch! I think the @ signs might be the wiki syntax for code styling. Do you know the appropriate syntax for the website?

rduivenvoorde:
double backticks around it: 

http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup

PR #140, opened 2014-02-18
Fix link to roadmap

m-kuhn:
Current link references the anchor incorrectly and therefore just jumps to page and not to paragraph


rduivenvoorde:
thanks, I just ran onto this one myself :-)

PR #167, opened 2014-05-23
Update screenshots.rst

anitagraser:
I think it's better to direct readers to the full Flickr group galleries if we don't have a more attractive way of integrating selected pictures on our website.

### DIFF
--- a/source/site/about/screenshots.rst
+++ b/source/site/about/screenshots.rst
@@ -7,43 +7,4 @@
 QGIS Screenshots
 ================
 
-Below are some screenshots from QGIS itself and a selection of Maps which are created with QGIS.
-
-
-These screenshots from the `QGIS screenshots - Flickr Group <http://www.flickr.com/groups/qgis-screenshots/>`_.
-
-:raw-html:`<p><a href="http://www.flickr.com/photos/madmanwoo/9693517656/" title="QGIS by Nathan Woodrow, on Flickr"><img src="http://farm3.staticflickr.com/2840/9693517656_d66efd5a87.jpg" width="500" height="396" alt="qgis"></a></p>`
-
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
hi anita, 
actually I just added a little script yesterday, which retrieved 30 small square thumbnails and put them beneaty the 5 big ones. Let me roll this out, show you and then discuss it

anitagraser:
Ah, looking forward to seeing that!

rduivenvoorde:
ok, build for english: http://qgis.org/en/site/about/screenshots.html

anitagraser:
Good work! Looks much better than before. Did you leave the big old pics in on purpose in this build?

rduivenvoorde:
yes. 
As said, actually I find those thumbnails too small, but because of all the different sizes of the maps AND the file size of bigger image I dare not to make them bigger.
Should I remove the big ones you think? Or should we just put 1 on top? If so which ones do you prefer?

To check sizes:
http://farm6.staticflickr.com/5575/14022747800_c5d31708e7_m.jpg
vs
http://farm3.staticflickr.com/2911/13858985554_29cae1e095_m.jpg

if you want to check out files with q.jpg are the square one, m.jpg is 204 px on longest side etc etc
                            // s - Specifies the small square size of 75x75 pixels.
                            // t - Specifies the thumbnail size with 100 pixels on the longest side.
                            // m - Specifies the medium size with 240 pixels on the longest side.
                            // z - Specifies the medium size with 640 pixels on the longest side.
                            // b - Specifies the large size with 1024 pixels on the longest side.
                            // q - square 

it would be neat if I find some javascript to make the m-size images square...

anitagraser:
I would just remove the big images. 

The thumbnails (using _q option) seem to be 150x150px but they are scaled down to 140x140px on the screenshot page. I can live with 150x150px size.

rduivenvoorde:
ok, removed the larger top images.

made the thumbnails span3 = 220 px

closing this one

PR #174, opened 2014-06-04
Update qugsnews.atom

anitagraser:


rduivenvoorde:
thanks, will be in after next build

PR #196, opened 2014-08-11
Fix broken link to INSTALL file on GitHub

icycle77:
Needed to be https not http.
Note this is my first ever github, git, and qgis change--bear with me if there's a mistake in process.


rduivenvoorde:
Thanks for letting us know! Will pull in this now, and be visible upon next build of website.

Good first try :-) 
Keep trying!
For us this is easier then an email...

PR #198, opened 2014-08-12
Update sponsorship.rst

claasleiner:


rduivenvoorde:
Hi, it's not clear to me what is the difference between the two pr's currently open, can you please explain if I should do both or not?
Bear with me, I do not want to check it by rebuilding the site every time, comments are also usefull to add to a PR. 
Another thing: if I'm correct you are changing icon? Can you please also remove the icon then which is not used, and all this in one PR?
If this is problematic, I will do, but as said, for us it is more efficient if we can see (or being told) in one glance what the pr is about to do.
thanks

claasleiner:
Hi there,
Unfortunately I do not know myself so well with github. I have now cleared the a PR. As the image file is no longer used to delete, I do not know.
With many thanks and best regards,
Claas Leiner

claasleiner:
Hi there,

Unfortunately I do not know myself so well with github. I have now 
cleared the a PR. As the image file is no longer used to delete, I do 
not know.

With many thanks and best regards,

## Claas Leiner

---

Dr.-Ing. Claas Leiner

Geodatenservice, Kartenwerkstatt &
GIS-Schule

Wilhelmshöher Allee 304 E
34131 Kassel
Tel. 0561/56013445 oder 0561/804-7196
Mobil: 01525/3876673
claas.leiner@eschenlaub.de

## http://www.eschenlaub.de

PR #208, opened 2014-10-10
Direct the get involved/core development link to the appropriate place

m-kuhn:
The current link points to issue reporting...


rduivenvoorde:
some time ago this pointed to this url
http://qgis.org/en/site/getinvolved/development/index.html#development
which at that moment was on top of the page.

The idea was that the development section should have been fleshed out more,
so a page about c++ development, python/plugin dev, doc writing etc etc.
But at this moment all (mostly about creating an issue) is put in one page.

I'm not sure if pointint to the INSTALL is really helpfull in this case

m-kuhn:
Go for that one then. There's a link to INSTALL just there so that should be fine.

PR #225, opened 2014-11-03
My first attempt with Farsi translation

saberraz:
Hopefully it won't blow up the website!


rduivenvoorde:
thanks saber, please do translations from now on via transifex.com:

https://www.transifex.com/projects/p/qgis-website/language/fa/

and I tested a build and all seems fine :-)

saberraz:
Hi Richard,

Thanks for pointing me to the right direction. I will use Transfix for future edits.

PR #238, opened 2014-12-18
made documentation more prominent

anitagraser:
A dedicated section for the training material seems a bit overkill. I changed the title to Documentation and made the link to the training material index less prominent. I reordered the sections because I think that the link to the documentation should follow right after the download.

### DIFF
--- a/themes/qgis-theme/forusers_index.html
+++ b/themes/qgis-theme/forusers_index.html
@@ -28,10 +28,10 @@
                             <a href="#download">{{ _('Download') }}</a>
                         </li>
                         <li>
-                            <a href="#plugins">{{ _('QGIS Plugins') }}</a>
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
Hi Anita, this is not our 'own' training stuff IN the documentation. It is training material provided by universities or third parties. There has been a discussion about it on the lists, and the outcome was that I would create 'landing'-page for that external material. Some authors thought to write pages theirselves to describe for example the curriculum, or how to use the material.

Not sure to call this 'Documenation', as I think we should keep that as a 'reserved' word for our own sutff

anitagraser:
I'm aware that this is external material. I think the title "training material index" does not convey the information that this is external material. Imho, it's confusing and not clear what to find under such a heading.

PR #242, opened 2015-02-08
Change Line 100

balkanick:
I've explained the issue after the dash, but the change I propose is this:
from "sudo apt-key add" to :
         "sudo apt-key adv"
This change is vital, because to my experience it actually adds the key.


elpaso:
I think the correct command is 
sudo apt-key add 
and not "adv"

Issue #169, opened 2014-05-23
Features page content is off topic / would be duplicate of docs


anitagraser:
The features page http://qgis.org/en/site/about/features.html#qgis-desktop actually covers the different applications developed within the qgis project but not qgis features.

If the features page would cover what the title suggests, it would duplicate http://docs.qgis.org/2.2/en/docs/user_manual/preamble/features.html. Maybe it should be emptied and just link to the docs.

m-kuhn:
How about moving the page about/features.html to something like about/products.html?

anitagraser:
+1 for moving. Not sure if "products" is the best term, maybe "applications"? The desktop application, server application, ...?

PR #255, opened 2015-04-09
Update and rename features.rst to applications.rst

anitagraser:
as discussed in #169


elpaso:
I think that the index must be updated too in order to retrieve the file with the changed name:
https://raw.githubusercontent.com/qgis/QGIS-Website/master/source/site/about/index.rst

can you please change the index too?

rduivenvoorde:
agreed with @elpaso renaming this file means that we will have a new resource in transifex, AND that all current translation will be invalidated.
Best to leave the filename as is plz, and only change the content

anitagraser:
If we can leave the file name but have the new title in the menu that would be fine with me. Is that the case though?

rduivenvoorde:
yep, please do

anitagraser:
Is it ok now?

anitagraser:
@rduivenvoorde: is there anything else that needs doing?

rduivenvoorde:
sorry for taking so long.... thanks!

PR #258, opened 2015-05-03
remove broken and incomplete case study from source tree

alexbruy:
This case study is incomplete and not included in toctree


dassau:
Hi,

please do not remove it. I started to add the case study but haven't found time, to finish yet.

alexbruy:
Ok, thanks for clarification

PR #280, opened 2015-09-25
update release version and add year to dates

3nids:


rduivenvoorde:
@3nids sorry for not taking care of this one... I see current version has even more info then yours. (and this one has merge problems...) so closing this one.

Thanks

PR #290, opened 2015-11-24
replace unused markup with explanatory comment

michaelkirk:
I think this is a little clearer, I see this pattern used in other pages (e.g. https://github.com/qgis/QGIS-Website/blob/master/source/site/index.rst)


rduivenvoorde:
@michaelkirk yep, you are right, it is replaced by a html/template ... this is old stuff.

thanks for cleaning up!

PR #293, opened 2015-11-25
remove toplevel "root" link in TOC.

michaelkirk:
People know what navigation is, we don't have to label it. Plus there's
a link to root at the top left. This just removes a bit of clutter.

Desktop Before vs. After

screen shot 2015-11-24 at 4 14 56 pm

screen shot 2015-11-24 at 4 14 36 pm

Mobile Before vs. After

screen shot 2015-11-24 at 4 31 05 pm
screen shot 2015-11-24 at 4 31 47 pm


rduivenvoorde:
Yep, you are right, sorry for taking me so long to look at this :-(

PR #302, opened 2015-12-19
Reorganizes the Development sectionTOC and content (1)

DelazJ:
fixes #294 and fixes qgis/QGIS-Documentation#566
This PR is **one of the two proposed to fix the issue**. See #303 for the other option. 

With this PR, This is how new site is structured:
website6
website2
website

It reorders the TOC  and the content of "development" section in "Get Involved" chapter: 
- Only the selected item in the TOC displays its children in the TOC.
- fixes the artefacts with the title style of "add grass tools" section
- groups c++ and python plugins sections together to avoid intermediate links that adds clicks to clicks
- splits into sections the "developer's guide for qgis" section to help the user easily reach the chapter he's looking for (keeping in mind that e.g, git section concerns not only QGIS coders but also documenters, a section can be added for ReST...)
- moves the "credits for contribution" section into "developer's guide for qgis" section
- the reason why there is a second PR is that I wanted to preserve as far as possible the current translations. Indeed, I think [Development section](http://qgis.org/en/site/getinvolved/development/index.html) is an old section that may have been widely translated so:
  - in this PR, i don't care about that and just split the text into **two new files : "Road Map" and "Bug, features and Issues"** which makes the chapter more file-modularized, and a first page with a more expanded TOC
  - in PR #303, I just "move" the text from the /development/index.rst file to a new file because I couldn't fix the TOC, keeping the content inside. I might miss something; The hope is that there could be an easier script to move their translations and not ask translators to redo it.

So questions are:
- Is it possible to easily move translations before pushing the changes to transifex ? if it isn't possible, don't care about #303.  however, if it's possible:
  - do we opt to keep these translations at the cost of a imho better structured document as proposed in #302? 
  - @rduivenvoorde do you know if Transifex still suggest translated text when the original file is split into new files (hence removed)?

PR #303, opened 2015-12-19
Reorganizes the Development sectionTOC and content (2)

### LINKED ISSUE #294: development nav hierarchy: dev guide should be up a level
I'm still learning the Sphinx/RST vocabulary, so forgive me, but there is a problem with the Development navigation item.

screen shot 2015-11-24 at 6 09 02 pm

I don't think the development guide should be showing up as a subitem of  "Credits for Contribution"

There could be a bigger refactoring in order, but the simplest direct thing that should be done, I think, is move the development guide from to "Get Involved / Development" > "Development" > "Guide"

### LINKED ISSUE #566: Some errors in QGIS-Development page
In [QGIS-Development](http://qgis.org/fr/site/getinvolved/development/index.html#qgis-development) page, I find the TOC weird:
- a problem with headers (about GRASS) : some subsections are shown like section (unless there are too many sections)
- there is a TOC shown at the right side but it doesn't correspond to the content of the page
- Is there the right place to put such informations about GRASS (too much technical)?  May be we should add this in the Documentation, after Processing features chapter (how to configure and "use" GRASS tools). C++ and Python plugins sections in QGIS-Development page just show links where to get more information. it should be same thing for GRASS I think.

DelazJ:
fixes #294 and fixes qgis/QGIS-Documentation#566
This PR is **one of the two proposed to fix the issue**. See #302 for the other option. 

With this PR, This is how new site is structured:
website5
website4
website3

It reorders the TOC  and the content of "development" section in "Get Involved" chapter: 
- Only the selected item in the TOC display its children in the TOC.
- fixes the artefacts with the title style of "add grass tools" section
- groups c++ and python plugins sections together to avoid intermediate links that adds clicks to clicks
- splits into sections the "developer's guide for qgis" section to help the user easily reaches the chapter he's looking for (keeping in mind that e.g, git section concerns not only QGIS coders but also documenters, a section can be added for ReST...)
- moves the "credits for contribution" section into "developer's guide for qgis" section
- the reason why there is a second PR is that I wanted to preserve as far as possible the current translations. Indeed, I think [Development section](http://qgis.org/en/site/getinvolved/development/index.html) is an old section that may have been widely translated so:
  - in this PR, I just "move" the text from the /development/index.rst file to a new file because I couldn't fix the TOC, keeping the content inside. I might miss something; The hope is that there could be an easier script to move their translations and not ask translators to redo it.
  - in PR #302, i don't care about that and just split the text into **two new files : "Road Map" and "Bug, features and Issues"** which makes the chapter more file-modularized, and a first page with a more expanded TOC

So questions are:
- Is it possible to easily move translations before pushing the changes to transifex ? if it isn't possible, don't care about #303. However, if it's possible:
  - do we opt to keep these translations at the cost of a imho better structured document as proposed in #302? 
  - @rduivenvoorde do you know if Transifex still suggest translated text when the original file is split into new files (hence removed)?

NB: in this PR, there is an inconsistency (local issue?) with the title "Road map and issues Report" that become "Issues Report and Road Map" in the third picture (when "developers guide" title is selected)


rduivenvoorde:
so this one is obsolete now? Plz close if so, or let me know what we want to do with this one.

m-kuhn:
Superseded by #302

PR #309, opened 2015-12-30
Fix text appearance

DelazJ:
Specially in [checklist for 2.0](http://qgis.org/en/site/getinvolved/governance/release_checklists/release-2.0.html), notes are wrongly splitted from line to line


rduivenvoorde:
Thanks!

PR #320, opened 2016-03-01
fix build warnings in italy_rome

DelazJ:
fix inline character of the main title
fix numbered list character, indentation and spaces
also break long lines


yjacolin:
LGTM, except conflicts :)

DelazJ:
see #321

PR #316, opened 2016-02-09
Fix minor alignment issue in Past Bronze Sponsors list

DelazJ:

PR #338, opened 2016-04-17
Fix book links

### LINKED ISSUE #328: Book recommendations links are broken.
On the coding standards page, all book recommendation links are broken.

Instead of navigating to the listed book, the user is redirected to http://www.informit.com/imprint/index.aspx?st=61085 which is not relevant.

DelazJ:
Should fix #328 

Just wonder: do we have some 'ethics' rules on website, e.g editor or seller, to link to?

### DIFF
--- a/source/site/getinvolved/development/qgisdevelopersguide/codingstandards.rst
+++ b/source/site/getinvolved/development/qgisdevelopersguide/codingstandards.rst
@@ -356,13 +356,13 @@ Book recommendations
 ....................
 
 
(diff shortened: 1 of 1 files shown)


DelazJ:
@andreasneumann Commit 4fe1967 has broken [Travis build](https://travis-ci.org/qgis/QGIS-Website/builds). Given that, if I'm not wrong (https://github.com/qgis/QGIS-Website/pull/316#commitcomment-16438797), there might be an automatic way to build the list, would you mind fixing it, please? (Doing it manually is just time consuming and can't be in my todo list).
Thanks

PR #341, opened 2016-04-19
Notify documentors webhook

### LINKED ISSUE #317: Adding tags to new commit to have it create issues in QGIS-Documentation
There are some instructions for developers in http://qgis.org/fr/site/getinvolved/development/qgisdevelopersguide/git.html. It might be interesting to let them know that there are some tags they can use while commiting to create issues list in Doc, Changelog...

DelazJ:
fix #317

### DIFF
--- a/source/site/getinvolved/development/qgisdevelopersguide/git.rst
+++ b/source/site/getinvolved/development/qgisdevelopersguide/git.rst
@@ -258,6 +258,21 @@ Best practice for creating a pull request
 * Attention: After a rebase you need to ``git push -f`` to your forked repo. 
   **CORE DEVS: DO NOT DO THIS ON THE QGIS PUBLIC REPOSITORY!**
 
+Special labels to notify documentors
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Besides common tags you can add to classify your PR, there are special ones
+you can use to automatically generate issue reports in QGIS-Documentation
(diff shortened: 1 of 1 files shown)


source/site/getinvolved/development/qgisdevelopersguide/git.rst
```diff
@@ -258,6 +258,21 @@ Best practice for creating a pull request
 * Attention: After a rebase you need to ``git push -f`` to your forked repo. 
   **CORE DEVS: DO NOT DO THIS ON THE QGIS PUBLIC REPOSITORY!**
 
+Special labels to notify documentors
```
DelazJ, line 261: Are only the manuals that benefit from tagging a commit or does the changelog also need them (in which case, which tags and how does it work)?

yjacolin, line 261: Changelog is currently manual. Only documentation issue are created with these labels.

rduivenvoorde:
@DelazJ originally the [FEATURE]-tag in the commit message in was used to be able to filter out the new features from all commits. To be used for the changelog and release texts.

We still do that, but now also have this issue-creating mechanism to make documenting new stuff easier, AND we created a new tag to instruct do writers to add documentation about old functionality which was changed or had additions.

Tags are:

[FEATURE ] and [NEEDS-DOCS]

https://github.com/qgis/QGIS-Sysadmin/blob/master/webhooks/github_feature_tracker.cgi#L429

PR #343, opened 2016-04-22
Add install instructions

DelazJ:
Discussion opened in the mailing-list at http://osgeo-org.1560.x6.nabble.com/Qgis-community-team-Moving-the-install-instructions-to-a-more-sexy-visible-coherent-and-translatableg-td5262483.html

Currently, when someone is looking for steps to compile a QGIS release, he'll will find them at http://htmlpreview.github.io/?https://github.com/qgis/QGIS/blob/master/doc/INSTALL.html

I personally don't like that page and from a non-english speaker and non-developer looking for a howto, I didn't like it not being translated. And there are pages from the translated website that directly link to this page (leaving a fully translated page to a specialized page totally in english from within the QGIS project doesn't really sound good to me).

Then, reading the introduction of this notice, you'll find assertions such as:
- "You can download this document as part of the QGIS 'User and Installation Guide' in HTML and PDF format via http://qgis.org."
- "Translations of this document can also be downloaded at the documentation area of the QGIS project at http://qgis.org."

AFAICS, this howto is available nowhere in the manuals and there's no translated version.

SO, does it make sense to move it to somewhere like
http://www.qgis.org/en/site/getinvolved/development/index.html?
It'll complete all other chapters on development matters, will be more visible, sexy and translatable.

The issues I can't figure out how they can be solved are:
- what will be the status of the current page, in the QGIS repo?
- how will update be managed?

### DIFF
--- a/source/site/getinvolved/development/qgisdevelopersguide/compile_qgis.rst
+++ b/source/site/getinvolved/development/qgisdevelopersguide/compile_qgis.rst
@@ -10,40 +10,40 @@ Last Change : *Friday April 15, 2016*
 Introduction
 ===============
 
-This document is the original installation guide of the described software 
-QGIS. The software and hardware descriptions named in this 
-document are in most cases registered trademarks and are therefore subject 
-to the legal requirements. QGIS is subject to the GNU General Public 
+This document is the original installation guide of the described software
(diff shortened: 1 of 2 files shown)


rduivenvoorde:
Closing some stale PR's

PR #353, opened 2016-06-01
Fix grammar error

cblument:
Start using QGIS ... reads better than
Started using QGIS ...

The other two highlights start with:
Shape the future ...
Find training ...


DelazJ:
Good catch! Thanks...

PR #364, opened 2016-07-11
Blocker priority no long exists in bug reporting

DelazJ:


yjacolin:
I don't understand the title of your PR and the content. Yo usaid priority doesn't exist anymore, so why keep the description?

DelazJ:
The option `Blocker` does no longer exist in the Priority drop-down list. It's replaced by `Severe/Regression`

yjacolin:
Thanks!

PR #366, opened 2016-07-17
Clean changelog 2.16

DelazJ:
Fix typos
Hyperlinks for some funders were wrongly placed
fix some indentation
add some directives for a better rendering
Add comments on doc guidelines improvement
break lines to 80 characters for a better reading and check

### DIFF
--- a/source/site/forusers/visualchangelog216/index.rst
+++ b/source/site/forusers/visualchangelog216/index.rst
@@ -6,32 +6,58 @@ Changelog for QGIS 2.16
 
 |image1|
 
-This is the log for the next release of QGIS - version 2.16.0 'Nødebo'. The Department of Geoscience and Natural Resource Management
-Forest and Landscape College in Nødebo were hosts to the First International QGIS conference and developer meeting in May 2015.
+This is the log for the next release of QGIS - version 2.16.0 'Nødebo'.
+The Department of Geoscience and Natural Resource Management
+Forest and Landscape College in Nødebo were hosts to the First International
+QGIS conference and developer meeting in May 2015.
 
 
 **This is not a Long Term Release**
 
-This release is not designated as a Long Term Release (LTR). Users wishing to have a version of QGIS which does not change and receives bug fixes for at least 1 year are invited to use the current LTR release 2.14 (which now becomes our official LTR with the release of QGIS 2.16).
+This release is not designated as a Long Term Release (LTR). Users wishing to
+have a version of QGIS which does not change and receives bug fixes for at least
+1 year are invited to use the current LTR release 2.14 (which now becomes our
+official LTR with the release of QGIS 2.16).
 
 **New Features in QGIS 2.16 'Nødebo'**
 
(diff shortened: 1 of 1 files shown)


andreasneumann:
@DelazJ @timlinux  - thank you for these improvements. But the changes should go directly into the source at http://changelog.qgis.org/en/qgis/version/2.16.0/ and not to a copy. Otherwise the source still contains the problems.

The reason the links are partially misplaced, is that the interface only allowed for one sponsor URL, not for several ones. This should also be fixed in the source application (Projecta) - not in the copies.

DelazJ:
@andreasneumann from what i've seen from previous releases, once the changelog is moved to qgis.org, changelog.qgis.org is no longer used. The source of the changelog becomes the one in this repo and thus the one to be improved is this one. The link to advertise (and translate) is not http://changelog.qgis.org/en/qgis/version/2.16.0/ but http://qgis.org/en/site/forusers/visualchangelog216/index.html if i'm not wrong.
changelog.qgis.org is just to draw the changelog, giving to anyone write access. Then as an unofficial link and not yet finished document, there's no risk to have user being puzzled/aware of that page.
Then I'm not sure there's a direct update process between changelog.qgis.org and qgis.org, so only one could be maintained.
@rduivenvoorde ?

DelazJ:
I'm unable to find how to solve https://travis-ci.org/qgis/QGIS-Website/builds/145358896 any help is welcome

rduivenvoorde:
you should NOT break lines to 80 characters as that breaks rst syntax!

Please let me know what other problems there are, and I'll try to fix them.

I'll close this PR for now

DelazJ:
@rduivenvoorde how does breaking lines breaks rst syntax. That's weird. We've often done this in the doc repo without breaking anything. Afaics, rst doesn't matter the length of the lines....
@yjacolin?
About the errors in the file, give a look to the second commit in the PR; they are typos, wrong indentation, weird url linked to more than the concerned funder and few improvement for reading I add. I'm not on a computer otherwise I'd have sent you direct link but all is already in the commit mentioned beforehand...

rduivenvoorde:
@DelazJ the export from projecta also broke because of mostly breaking long(!) url's and those parts of text that had a | in them.
But I will review the whole PR then by hand here...

rduivenvoorde:
@DelazJ ok I see you problem now... you 'fixed' all url's by removing the second underline in the (often identical) url's like those 'developed by ...', but those double __ is actually right :-)  It means an 'anonymous' url... see http://stackoverflow.com/questions/5464627/how-to-have-same-text-in-two-links-with-restructured-text

So actually 1f6641e is wrong and should be removed from this pr. 

What do you prefer? I think it is better to redo the other commits and do a new PR?

DelazJ:
Revert the `__` removal in link

DelazJ:
@rduivenvoorde I rebase the branch and fixed the commit according to your previous comment. It might be good now.

DelazJ:
ping :-)

PR #385, opened 2016-10-13
Updated ticket fieldnames

kgjenkins:
To match actual fieldnames in the "New issue" form.  Also changed mentions of "ticket" to "issue" to be consistent.

Also, "milestone" should get updated, but I'm not sure what it corresponds to...

### DIFF
--- a/source/site/getinvolved/development/bugreporting.rst
+++ b/source/site/getinvolved/development/bugreporting.rst
@@ -79,12 +79,12 @@ Steps
 ^^^^^
 
 To report a bug choose New Issue from the menu bar. Note: You can also request 
-an enhancement or submit a patch using the Ticket system.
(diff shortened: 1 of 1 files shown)


DelazJ:
@kgjenkins thanks for the fix (i have broken :( )
I suppose that `milestone` corresponds to `Target version` (_from a translated form_), the version in which you want your issue fixed. However IMO this has sense only when requesting a feature given that we, most of the time, want a bug fixed in the currently affected version, as soon as possible. But so is the form.

I wonder if "ticket" has to disappear from the description (note that it's the expression used in the right panel, with prepared requests). No strong opinion on this however.

rduivenvoorde:
Thanks, I'll pull and fix

PR #402, opened 2016-12-06
Added QGIS.ZA user group

timlinux:


yjacolin:
@timlinux some commit doesn't appear in the diff tabs. Did you rebase your branch or so? Could you confirm that you only want to add the South African QGIS User Group. Thanks.

yjacolin (approved):

PR #405, opened 2016-12-13
fixup some path

vmora:
@mhugent isn't the config of the server address missing ?

I tryied to add the yes.xml and a custom wms-XXX.xml froms as recomended here http://opengeospatial.github.io/teamengine/users.html but with no success.


DelazJ:
@mhugent are you ok with these changes?

PR #406, opened 2016-12-18
Move the Developers Guide from website to Documentation repository

DelazJ:
As discussed in https://github.com/qgis/QGIS-Documentation/issues/1566, this PR begins the move:
* remove the developers guide chapter
* add a comment about QGIS Core development, linking to the new placement

### What it doesn't do:

Move the [Adding GRASS tools](http://qgis.org/en/site/getinvolved/development/addinggrasstools.html). I'm not really a GRASS user so wonder:
* what does really this text refer to? is it about general coding skills (thus can be moved to the new place) or could it be put in the GRASS chapter of user manual or processsing grass section?
* is this descripton up to date?

Digging the history, I found that @anitagraser creates this file. Do you have some opinion on above?

### Things to take care of

* There's a pending PR at #405 we should think to merge in the new place (would be easy)
* **Translation** (see qgis/QGIS-Documentation#1585)


anitagraser:
I don't know if those are up to date. I didn't write them, I just recreated them based on some previous description ... might have been on the Wiki originally.

DelazJ:
Thanks @anitagraser

SrNetoChan:
Because website updates are automatic, to make sure we don't lose information while the new documentation is not released, I would keep this page in the website until the next documentation release.

rduivenvoorde:
Hi @DelazJ , I was cleaning up the pull requests here, is this PR still to make valid? Or should/can I close it?

DelazJ:
@rduivenvoorde Thanks for the cleanup. It's very welcome.

About the PR, hummm.. Somehow it needs to be done or something needs to be done! Currently (for 2 years actually), seems that nobody noticed it but we have two dev guidelines diverging (https://docs.qgis.org/testing/en/docs/developers_guide/ vs https://qgis.org/en/site/getinvolved/development/qgisdevelopersguide/index.html despite the warning on top of the latter link). And at some point THAT should be fixed.
Other than the wait for the (2.18?) doc release that makes the PR stall, there were some related questions that also refrained me:
1. what do we do with the grass section here? I had opened a discussion on the list but not sure of the outcome, if any
1. whether we keep the dev guidelines in the doc repo (the famous discussion about splitting), to avoid doing the move twice

I'll try to move forward this PR if I get time in next days.

(Not related to this PR) I was looking at the website yesterday and was telling myself that its tree would need to be refactored (see eg how heterogeneous are the items of "get involved/development" - a new header might be welcome and some items moved to another "umbrella") - my2cts.

DelazJ:
@rduivenvoorde rebased and new commits I found are pushed to doc repo (https://github.com/qgis/QGIS-Documentation/pull/3989)

PR #420, opened 2017-03-15
Adding 2.18 doc pages

DelazJ:
To be able to:
* use the Fix Me link on online 2.18 doc, and not always land on master branch;
* and easily compare the two docs when needed.

@rduivenvoorde could you do the other magics behind the scene to make it available? Thanks


DelazJ:
done by @rduivenvoorde meanwhile

PR #426, opened 2017-04-08
consolidate almost empty meetings page to reduce number of resources

### LINKED ISSUE #395: Future of getinvolved/meetings section
This section contains pages for almost all developer and user meetings. But most of the pages are empty, does it worth to keep them? Maybe we can have single page with links to blogposts/reportbacks?

alexbruy:
Fixes #395


source/site/getinvolved/meetings/developer/index.rst
```diff
@@ -2,23 +2,198 @@
...
+
+Many thanks to our sponsors!
+
+ - University of Copenhagen
+ - http://www.ku.dk/english
+ - https://www.lifa.dk/
+ - http://www.dge.dk/
+ - http://www.gisgroup.dk/
+ - http://www.septima.dk/
+ - OSGeo
```
DelazJ: This list (and followers) should not be indented

DelazJ:
Thanks @alexbruy For sure this makes these chapters more readable, consistent and somehow cohesive

I just wonder if we should also remove files mention from `.tx/config file`: In the case merging this PR removes them also from Transifex, that would mean that translations corresponding to the texts you move will also be removed from Transifex, hence translators will need to rewrite/retranslate instead of simply pick the right text from Transifex suggestions.
Do I understand how things work @rduivenvoorde ? In which case we could wait some reasonable months and remove them later...

rduivenvoorde:
Thanks @alexbruy !!

@DelazJ no need to remove them from .tx/config, whenever we update the transifex strings we run the scripts/create_transifex_resources.sh and all (from github/filesystem) removed files will not show up in .tx/config anymore (before running the script I remove all current content to be sure).

Files on transifex are not removed untill we remove the resource one by one by hand. And yes, it is better to give the translators some time to move translations.
So when we think translators are ready, just hand me a list of files, and I remove them.

alexbruy:
@DelazJ, @rduivenvoorde personally I think that we can safely remove all resources from Transifex as these pages were empty.

PR #445, opened 2017-06-17
Updated release schedule

m-kuhn:
The current roadmap for 3.2 and up is based on a [miscalculation in an email](https://lists.osgeo.org/pipermail/qgis-developer/2017-February/047113.html) which was later on [updated](https://lists.osgeo.org/pipermail/qgis-developer/2017-February/047121.html).

based on https://gist.github.com/jef-n/796ec9c7bbc2ee22032a4d394b69170f#file-gistfile1-txt-L306-L310


m-kuhn:
I am not sure about the values in the last column

m-kuhn:
@pcav you have written the announcement email, did I get this right?

pcav:
Not that I know of :)

m-kuhn:
@pcav click the link in the topmost message, your name is there ;)

jef-n:
2.18 should be the LTR for at least 12 months.  3.2 should be the next LTR.

m-kuhn:
3.2 was already meant to be the next LTR in the proposed table. Good call for the LTR duration.
What do you think about this?

| Release | Long Term Release | Date       |
|---------|-------------------|------------|
| 3.2.3   | 2.18.19           | 20.04.2018 |
| 3.2.4   | 2.18.20           | 18.05.2018 |
| 3.4.0   | 3.2.5             | 29.06.2018 |
| 3.4.1   | 3.2.6             | 27.07.2018 |

jef-n:
The release of 3.2 must be later than 19.5.2018 to make 2.18 last a year.  The current schedule puts it to 29.06.2018 for that reason.

m-kuhn:
That's pretty much what's done here. 3.2.5 is announced as LTR in June next year.

3.2.0 is a regular release just like 2.18.0, 2.18.1 ... were until May this year. Or just like 2.14.0 was already released when 2.8 was only LTR for 8 months.

jef-n:
3.2.0 is the LTR - not 3.2.5 - that's just when the LTR replaces the earlier LTR in the ltr package repositories.

m-kuhn:
> that's just when the LTR replaces the earlier LTR in the ltr package repositories

... which in turn means that 2.18 will be available in this repository until this day?

m-kuhn:
Sorry for insisting. What I just don't understand is, why there is no release planned (LTR or not) 4 months after 3.0 is released.

DelazJ:
Considering that I've raised the same questions few months ago http://osgeo-org.1560.x6.nabble.com/Plans-for-next-QGIS-releases-tp5308601p5315313.html maybe it needs some public clarification(and share) in ML (once for all) on how things are processed:
- when do we call a release LTR: at the release of the branch (or whatever it's called at that time) or four months later when it replaces the previous LTR? (communication consideration)
- how/when a release become LTR? (technical consideration) 
- why 3.2 would not released at the beginning of 2018, ie 4 months after 2.18?
- ... (_other things people might need to know)_

my 2cts

jef-n:
I don't know why 3.2 should be the next LTR - that wasn't discussed at the time - but there seemed to be agreement on that.   But that moves it 4 months ahead (because it has to be at least 1y after the 2.18 LTR; maybe to give 3.0 four more month to settle).   Otherwise it would have been a regular release and 3.4 would become the next LTR - which would have been my preference - but I didn't want to argue on that either way - because I'm currently not really involved in the actual feature work (so if the doers request 4 more month long ahead - ok with me).

m-kuhn:
> I don't know why 3.2 should be the next LTR

IIRC because 3.0 was out of question and it shouldn't be too long after.

I'm still having a hard time to understand this. I'll shortly outline what to me appears to have been the general plan of action in the past.

A subsequent LTR will ...

* ... be released as .0 release a year or more after the previous one (2.14.0 a year after 2.8.0)
* ... replace the the preceding LTR in the ltr package repos a year or more after the previous one (2.14.4 a year after 2.8.4)

Now if 3.2.0 is released in Jan (or Feb) and 3.2.4 (or 3.2.5) replaces 2.18.19 in the ltr package repos in May/June, both of the above rules are followed.

2.18 will receive a year of patch releases in the ltr package repos and way more patches than any other branch over its whole life. So to me everything seems to be fine with this scenario (and looking at the quote from the IRC conf "release 3.2 as next LTR in release 3.0 + 4 Months" seems to be in line with expectations)

What am I missing?

jef-n:
>   I don't know why 3.2 should be the next LTR
> IIRC because 3.0 was out of question and it shouldn't be too long after.

That doesn't make sense to me timewise - skipping a release doesn't brings the next LTR any closer to 3.0.    If we did a regular 3.2 after four month and a 3.4 LTR after four more month we'd end up with the same time and it would have been the regular schedule.

> What am I missing?

Nothing, I guess.  We just decided against the regular schedule.   I didn't question why at the time and am still fine with it - but I wouldn't mind the regular schedule either.

BTW for the year I count from release to release and ignore the extra four month of repository overlap.

m-kuhn:
> That doesn't make sense to me timewise - skipping a release doesn't brings the next LTR any closer to 3.0. If we did a regular 3.2 after four month and a 3.4 LTR after four more month we'd end up with the same time and it would have been the regular schedule.

I don't find any reference to the request for a double cycle for 3.2 nor would I be aware of anyone asking for that (but I may be mistaken). Just this phrase: "release 3.2 as next LTR in release 3.0 + 4 Months (eta June 2018)" which isn't clear to interpret but might make sense taking into account the mixup in terminology mentioned by @DelazJ

> BTW for the year I count from release to release and ignore the extra four month of repository overlap.

no problem then, 2.18.0 + 1 year is end of 2017, 3.2 will not be out by then ;)

jef-n:
> I don't find any reference to the request for a double cycle for 3.2 nor would I be aware of anyone asking for that (but I may be mistaken). Just this phrase: "release 3.2 as next LTR in release 3.0 + 4 Months (eta June 2018)" which isn't clear to interpret but might make sense taking into account the mixup in terminology mentioned by @DelazJ

Well, release 3.2 June 2018 was clear enough to me.  That's what we decided and what has be on display for 4 months

Did we move the feature freeze of 3.0 after the discussion?  On the roadmap it's on 18.8., but the log states it should be in july.

@timlinux @nyalldawson @pcav @m-kuhn @haubourg

m-kuhn:
My theory is June was based on 3.0 + 4 months dev (release 3.2.0) + 4 months cooldown (3.2.4) before packaging as ltr + a rounding error. Would be great if others express their thoughts :)

m-kuhn:
Ping @timlinux @nyalldawson @pcav @haubourg @NathanW2 @3nids @andreasneumann 

Does anyone have an idea why we ship 3.2.0 *9 months* after the release of 3.0 (and not ship a 3.2.0 release *4 months* after the release of 3.0)?

That's what the roadmap currently states but I could not find any reason for this gap.

3nids:
I am not aware of any discussion to increase the delay between 3.0 and 3.2.

To me, the schedule @m-kuhn detailed makes much sense:

Date | current | LTR
------ | --------- | -----
2017.06 | 2.18  | 2.18 
2017.10 | 3.0  | 2.18 
2018.02 | 3.2  | 2.18 
2018.06 | 3.4  | 3.2 

It does respect all what I heard in the disucssions: 1 year of LTR point releases, 4 months cylces.

Any objective reason why not to stick to it?

jef-n:
> Any objective reason why not to stick to it?

Stick to what was decided (ie. 3.2 in June 2018)?

3nids:
Any reference to the decision? 
As pointed out I wonder if it's not a misunderstanding of when a version becomes LTR (when it's out or when the previous gets retired).

jef-n:
@3nids in the links of the first comment.

DelazJ:
I'm sorry but this begins to look like an endless (and unfruitful?) discussion (given that every body seems to refer to the same source, with different interpretations).
If there's no more voice from people both Juergen and Matthias pinged, better move the discussion to the psc and request an official position on what should be, once for all...
my 2cts

m-kuhn:
> @3nids in the links of the first comment.

... which state that

 * 3.2 should be released in June 2018
 * 3.0 should be released in September 2017
 * 3.2 should be released 4 months after 3.0

really?

m-kuhn:
@jef-n are the results of this discussion really clear and meaningful to you?

I don't want to raise a psc discussion without having you as release manager at least acknowledging that there might have been a communication problem.

Personnally I think this roadmap has many disadvantages like almost nobody testing 3.2 before June and a production stable release around 3.2.4 not before autumn 2018. On the other hand I cannot see any advantage of it.

@jef-n your call: if it's really all clear as the sky to you, please close this PR. If you agree (partially) with these concerns let me know and I can write a mail to the PSC list.

Thanks a lot

timlinux:
Sorry for joining the thread late @m-kuhn @jef-n @DelazJ and @3nids 

Can I propose that we put this on the agenda for coming Monday's PSC meeting (8pm CET) and that you join us to discuss it there?

Please distill the discussion to a simple tabular breakdown for each option so that the variances are easily digestible to use that are trying to pick up the threads of conversation?

jef-n:
@timlinux comment on https://lists.osgeo.org/pipermail/qgis-psc/2017-June/005398.html please.

DelazJ:
I guess we could close this PR now

PR #458, opened 2017-08-12
typos corrected

DiGro:
line 147 :  help   should be  :  held
line 148 :  Lison  should be  :  Lisbon
line 169 :  help   should be  :  held


rduivenvoorde:
thanks

PR #464, opened 2017-09-06
clarify that ubuntugis-unstable is what we build against

mbernasocchi:
made more clear that  the ubuntugis based repos need ubuntugis-unstable PPA


yjacolin (changes requested):
@mbernasocchi would you mind to rebase your branch?

mbernasocchi:
@yjacolin done

yjacolin:
@mbernasocchi I think you need to change the table a little bitt as you add more letter in some rows!

yjacolin:
Thanks @mbernasocchi :)

PR #467, opened 2017-09-13
Warning Ubuntu version compatibility with master

m-kuhn:
Specify where to get master builds for Ubuntu 16.04.

The version currently offered by debian-nightly is from January 7. As people (hopefully) start testing QGIS 2.99 more, we don't want them to report issues against this version. Still it's quite likely that people use this because it's (IIRC the latest) LTR and people don't read footnotes ;)

In addition to this, I would propose to remove the package from the 16.04 repository, @jef-n what do you think?


jef-n:
Packages removed - but people should still read footnotes.

mbernasocchi:
@jef-n should I close this PR then?

PR #468, opened 2017-09-16
Clarified trademark usage

anitagraser:


timlinux:
That looks good @anitagraser thanks. Perhaps you can add a short sentence like:

"When referring to QGIS please do not undermine the validity of the main QGIS project. We prefer that you refer to our software packages as 'Official Packages' and not 'Community Releases' or similar. If you create your own packages, please make it clear that these are derivative packages and not created or endorsed by the QGIS project."

anitagraser:
Thanks @timlinux! I think that's an important note, even if it's more of a "marketing guideline" than strictly a trademark issue.

pcav:
Agreed, thanks for taking care of it.
I just wonder if it wouldn't be better to just choose one formula and stick to it.
XX powered by QGIS seems the best to me.

anitagraser:
I think it would be good to announce the update on the developer mailing list and collect some feedback from the community.

PR #473, opened 2017-10-04
Emphasize the LTR releases

DelazJ:
I still see that despite all the text in http://qgis.org/fr/site/getinvolved/development/roadmap.html#release-schedule people still think that QGIS 2.18 is not yet a LTR (see QGIS-tr discussion) and, I guess, will be LTR when it's shown in the Long-Term Repo column (hence in december). People just focus on the table.
Given that, if i well understood, from a user point of view we currently have 2 LTRs, do you think that reorganizing the table this way could do the trick? Focusing the table on releases (and not repo) and show that there are cohabitation at some point between LTRs and that in dec only 2.18 will remain LTR and 2.14 dropped and later 2.18 will live aside 3.4 as LTR before being dropped... and so on...
@jef-n @m-kuhn @rduivenvoorde 
image


source/site/getinvolved/development/schedule.inc
```diff
@@ -1,33 +1,33 @@
...
+========== ======= ================= ===== ========== ==== =====
+Event      Latest  Long-Term         FF    Date       Week Weeks
+                   Release                            #
+========== ======= ================= ===== ========== ==== =====
+PR         2.18.4  2.14.12                 2017-02-24 9     4
+PR         2.18.5  2.14.13                 2017-03-24 13    2
+EPR        2.18.6                          2017-04-07 15    2
+PR         2.18.7  2.14.14                 2017-04-21 17    4
+**LTR**    2.18.8  2.18.8/2.14.15          2017-05-19 21    1
+EPR        2.18.9                          2017-05-26 22    4
```
DelazJ, line 10: I don't know how to handle this line

m-kuhn:
I don't think that helps anything (the target audience doesn't look at this table).

I'd **a lot** prefer to adjust the page here:
https://www.qgis.org/en/site/forusers/download.html

1. 2.18 on top
2. Label 2.18 with "Corporate Users" (instead of "New Users")
3. Label 2.18 with "LTR"

At the moment, there is no reason for anyone to roll out 2.14 on a new installation. Except for those who already have an infrastructure running on 2.14, and they already know what to look for.

In the future:

1. Replace the term "New Users" in current release with "Latest Features"
2. 3.4.0-3.4.3, label it with "Upcoming LTR"

jef-n:
Don't make manual adjustments to the schedule.inc - it's produced by scripts/update-schedule.py

jef-n:
@m-kuhn see a6016ba8c

DelazJ:
> Don't make manual adjustments to the schedule.inc - it's produced by scripts/update-schedule.py

OK @jef-n. In that case, I'm not sure I can fix that.
Anyway, this is a proposal for reorganizing that table. If ever it makes sense...
> @m-kuhn see a6016ba

Should I understand that I waisted my time because of your direct commit and no review/comment on #474 ?

jef-n:
> Should I understand that I waisted my time because of your direct commit and no review/comment on #474 ?

Didn't notice #474.  Could have been part of this or mentioned here earlier.

PR #474, opened 2017-10-04
Refine the download page instructions

DelazJ:
This is imho a complementary step to #473 . Changes include:
- a clear statement about which version is currently the LTR
- stop opposition between long-term release and latest release; they are all standalone binaries
- osgeo4w is not for advanced users (this worked like a barrier for me for years); it's just the best install tool 😄 that needs some explanation to cover why ltr can be available in lr repo.

@m-kuhn ?
image
image

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -1,8 +1,10 @@
 <section class="intro">
     <div class="container">
         <h2>{{ _('Download QGIS for your platform') }}</h2>
-        <p class="main">{{ _('The current version is QGIS %s \'%s\' and was released on %s.')|format( release, codename, releasedate.strftime('%d.%m.%Y') ) }}</p>
         <p class="main">{{ _('QGIS is available on Windows, MacOS X, Linux and Android.') }}</p>
+        <!-- Update the line below when the latest release is no longer the LTR:  replace version by ltrversion -->
+	<p class="main">{{ _('The current long-term release (for corporate users) is QGIS %s \'%s\'.')|format( version, codename ) }}</p>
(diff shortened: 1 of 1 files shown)


m-kuhn:
Do you have a screen shot?

DelazJ:
Will build the site tonight...

m-kuhn:
I'm not sure what exactly all the placeholders will expand to.

IMO, at this moment right now we do not need to introduce a text that works for the next 5 years, we will only need to introduce a text that works for the next 2 months.

DelazJ:
Screenshots added

jef-n:
Doesn't mention why 2.14 is still downloadable or why it's the only one of the other available standalones that is still mentioned.   But I too never really got why the standalone installers should be preferable over the OSGeo4W installer.  I hardly ever use the standalones.

m-kuhn:
Maybe an additional line on top "QGIS 2.14 LTR will be receive updates until [put EOL date here]" help?

anitagraser:
I'd also welcome putting OSGeo4W on top and making stand-alone installers less prominent.

m-kuhn:
As a new user who has no idea about packaging and thought "nice, let's try this QGIS thing", how am I going to know that QGIS is included in this 7-character-thing *OSGEO4W*? Me in such a situation would probably go for the standalone, because hey, I wanted QGIS, and I don't even know it's in there. Do you think the term "flexible QGIS installer" or similar could help here?

While at it, should we also put 64bits before the 32bit versions? I guess that's more common nowadays.

yjacolin:
I agree to put 64bits before the 32bit versions and I agree also that new users use QGIS standalone installer instead of OSGeo4W but I can't say that is because it is simpler to understand or in top first place. We can use something like "Custom QGIS 

My comment doesn't give too much help but I would like to clean up waiting PR and make a decision :)

jef-n:
30d16296 implements this (with some revisions - eg. leaving the labeling of LR/LTR standalones in place, osgeo4w package names).

PR #497, opened 2018-01-30
link to instructions to install qgis from source

andrewharvey:
As a user I wanted to install QGIS, went to this page on the website, but couldn't find the option to install QGIS from source, I think it will be useful to add this link for more advanced users who would prefer to install from source.

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -461,4 +461,4 @@ QGIS Testing warning
 Installing from Source
 ======================
 
-Refer to the [INSTALL guide](http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTALL.html) on how to build and install QGIS from source for the different platforms.
+Refer to the `INSTALL guide <http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTALL.html>` on how to build and install QGIS from source for the different platforms.
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -461,4 +461,4 @@ QGIS Testing warning
 Installing from Source
 ======================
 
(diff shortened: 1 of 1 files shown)


source/site/forusers/alldownloads.rst
```diff
@@ -458,3 +458,7 @@ QGIS Testing warning
    given time not work, or may do bad things to your data.
    Take care. You have been warned!
 
+Installing from Source
+======================
+
+To build and install QGIS from source see https://github.com/qgis/QGIS/blob/master/INSTALL
```
yjacolin: Could you use this link instead: http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTALL.html

yjacolin (changes requested):

source/site/forusers/alldownloads.rst
andrewharvey: sure, I've changed the link

yjacolin: thanks it could be even better if you can use something like 
```
[QGIS from source](http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTALL.html).
```
Which will be displayed like this:
[QGIS from source](http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTALL.html).
Sorry to forget this in my previous comment.

andrewharvey: I've used similar wording as the Sources tab at https://www.qgis.org/en/site/forusers/download.html. For some reason I never found that tab and only found the guide at https://www.qgis.org/en/site/forusers/alldownloads.html.

yjacolin:
@andrewharvey sorry I was wrong about the tags element. I updated your branch directly. I will merge it after the test. Thanks a lot!

yjacolin (approved):

PR #501, opened 2018-02-13
Add instructions on how to install QGIS Server on Debian

DelazJ:
in order to avoid duplicate in documentation (see https://github.com/qgis/QGIS-Documentation/issues/2246 and https://github.com/qgis/QGIS-Documentation/issues/2344)

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -224,10 +224,16 @@ following line::
 
  deb     http://ppa.launchpad.net/ubuntugis/ubuntugis-unstable/ubuntu *codename* main
 
-After that type::
+After that type the commands below to install QGIS::
 
  sudo apt-get update
  sudo apt-get install qgis python-qgis qgis-plugin-grass
+
+In case you would like to install QGIS Server (note that it's not a good practice to
+install both client and server applications on the same machine), type::
+
+ sudo apt-get update
+ sudo apt-get install qgis-server python-qgis
 
 .. note:: Please remove all the QGIS and GRASS packages you may have
    installed from other repositories before doing the update.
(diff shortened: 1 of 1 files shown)


source/site/forusers/alldownloads.rst
```diff
@@ -224,11 +224,17 @@ following line::
...
  deb     http://ppa.launchpad.net/ubuntugis/ubuntugis-unstable/ubuntu *codename* main
 
-After that type::
+After that type the commands below to install QGIS::
 
  sudo apt-get update
  sudo apt-get install qgis python-qgis qgis-plugin-grass
 
+In case you would like to install QGIS Server (note that it's not a good practice to
+install both client and server applications on the same machine), type::
```
yjacolin: why this is not a good practice to install both client and server applications on the same machine?

DelazJ: I don't know. I'm not a server user. This is an advice i read in some discussion on the lists (from different people). So if it's wrong, let me know and it will be removed

yjacolin: ok for my point of view, it is not,  I can't see any rational about this. But for production use, QGIS server need probably more ressource than for testing purpose but this is similar to every service (DB, MapServer or so).

elpaso: A server machine is typically (99.99% of the times) NOT a desktop machine, i.e. it is not *good practice* might be better said as *common practice*.

That said, I have a customer that does exactly that: he has installed both desktop and server component on the same machine, but as I said this is rather exceptional.

yjacolin: @elpaso I clearly understand that and agree with this :) so it is not "to install both client and server applications on the same machine" but rather "to install a server on a desktop machine".

elpaso: Well, the opposite is also true: "do not install a desktop on a server machine" :)

I think that everything boils down to this: we should not advice the (unexperienced) users to install server and desktop on the same machine, because that is not common practice.

If this is confusing, just remove the whole sentence.

PR #502, opened 2018-02-24
fix download link in carousel

noerw:
link was 404 due to relative `site`.


rduivenvoorde:
We are aware of this error, but we have a release3 branch with much more updates standing in line (and I will rebuild the sites immediately), as soon(!) as there are actual packages.

Current banner was put there too early...

rduivenvoorde:
FYI: this is the correct sphinx way to create links:

https://github.com/qgis/QGIS-Website/blob/release3/themes/qgis-theme/index.html#L48

noerw:
okay, thanks for the info.
just thought it would be good to fix a 404 link quickly

PR #515, opened 2018-04-12
Temporarily remove the wrong Q&A about domains

mbernasocchi:
removing this leaves "Examples of use requiring a permission: ........ If you use QGIS trademark in a domain name, with or without commercial intent". which is IMHO very strict for now but good until we vote a definitive scheme.


rduivenvoorde:
@mbernasocchi sorry Marco, did not see this PR
Did exactly the same and started a build, now visible already

andreasneumann:
Thanks!

mbernasocchi:
@rduivenvoorde no problem, the important thing is the paragraph is gone.

PR #516, opened 2018-04-16
Fix macOS version number

DelazJ:
Because version 3.0 is shown twice.
What could also be nice is to fix the line breaks around the python 3.6 hyperlink (and resize subtitles?)
image


kyngchaos:
Another change needed, I'll follow up with more fixes

PR #520, opened 2018-04-26
Donor Res.Gea Srl added

ghtmtt:


DelazJ:
Hi,
This commit broke the site build. Isn't the donor list scripted or something like that?

ghtmtt:
I get the error.. anyway your is a really good point. I could find anything (in this repo) where to add new donors so I did that manually. In the meanwhile I've reverted the commit: I get the error, so I should be able to correct it manually. @rduivenvoorde is there another correct way to add donors manually?

ghtmtt:
BTW: thanks @DelazJ for catching this ;)

andreasneumann:
I don't know why it broke:

@ghtmtt - usually you add new donors at the very end. A script will later sort donors automatically. Also make sure you always have a section of three, not 4. Maybe that broke the site rebuild?

As I told you - I will add donors manually who paid by wire transfer.

Did they request that they should be listed? I normally do it once every couple weeks.

ghtmtt:
@andreasneumann yes they sent me an email to be mentioned in the website. 

I saw that in the `donors.inc` all the names are sorted and that's because I added it not at the end (breaking so the build). Should I add it at the end or should I place it in the sorted place paying attention to have a list of 3 bullets?

andreasneumann:
@ghtmtt always at the very end.

And make sure there is always a block of three.

Don't worry about sorting. This will be done automatically.

ghtmtt:
ok done #522 let's wait for the check. @andreasneumann in the meanwhile can you merge the PR of the reverted commit? Thanks!

jef-n:
I guess this could also be fixed by donating something - the paypal responder script should also fix things - but only if you didn't already donate :)

andreasneumann:
;-) didn't Paolo write a "call for donors" just recently ;-) ?

PR #535, opened 2018-05-25
Fix training materials hierarchies

kinow:
It looks weird the current menu being

* Brazil
    - English
    - French
    - German

I think it should instead be

- Brazilian Portuguese
- English
- French
- German

Assuming too that it was intended to leave the languages in alphabetical order. Otherwise maybe English first as it has the largest number of materials, I think.

Cheers


rduivenvoorde:
@kinow you are right, that has the wrong header type.

Thanks!

I think we should start with english yes, and after that alphabetical. Can you change to that please?

kinow:
Thanks for the quick reply @rduivenvoorde 

>I think we should start with english yes, and after that alphabetical. Can you change to that please?

Sure thing. Done. Also made the break lines uniform between paragraphs.

rduivenvoorde:
Cool, Thanks!!!

PR #542, opened 2018-06-27
Clarify the Ubuntu installation instructions for new users

nickbearman:
I have added some clearer and (I hope) easier to follow instructions for new users. There is some tidying up to do, but I wanted to get people's views on this (see the comments in the PR).

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -99,6 +99,39 @@ cutting edge QGIS testing build (note the warning_).
 Debian/Ubuntu
 -------------
 
+## Easy Installation
+
+1. Open 'Ubuntu Software'.
+2. Search for 'QGIS Desktop'.
+3. Click 'install'.
+
+*This may install an older version, at time of writing this installs version 2.14.18. If you want a newer version, please see below.*
+
+## More Detailed Installation
+
+*This will install the latest stabe version (3.0.x Girona). If you want something different, please see below.*
+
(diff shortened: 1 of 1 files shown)


nickbearman:
Some more comments:
I think we need to make the instructions for Ubuntu users easier, as for people new to Ubuntu / Unix, things are quite confusing to get QGIS installed. 

The "Easy Installation" section currently doesn't work. This would require QGIS to be listed in the 'Software' program in Ubuntu - does anyone have experience of doing this / is this something we can do?

The "More Detailed Installation" does currently work, and takes the user through adding universe, and adding the https://qgis.org/debian repository. I don't think we should confuse people with multiple repositories here - just say "use this one".

Common Problems is fairly self explanatory, and I will add to this in due course. 

After I have currently just left the current instructions as they are. We can tidy this up if we go ahead with the new instructions above. 

Comments, thoughts, suggestions, etc. all welcome. 

Best wishes,
Nick.

reginaldc:
Hi Nick,

I have read the instructions for ubuntu that you propose. These are all fine except that half of the qgis users  want to install the 2.18.21 version (as on Osgeo4w in Windows).
Could you add the procedure to do this on Ubuntu?

Thanks for this nice work,

elpaso:
@reginaldc just curious: can you share the source of your statistics about the 50% of users willing to install 2.18.21 version? 
Do you have a confidence interval for your figures?

reginaldc:
No, it's just a rough estimation based on the use of qgis in the workplace and in other municipalities I know.

rduivenvoorde:
@nickbearman thanks for you work, sorry for not pulling earlier.

PR #544, opened 2018-07-04
Fix language names

sikmir:
Fix some wrong native language names:
- (Finnish) _Suomalainen_ ➟ **Suomi**.
_Suomalainen_ (adjective) means "relating to the Finnish language or Finland", but language name is **Suomi** (noun).
- (Lithuanian) _Lietuvos_ ➟ **Lietuvių**.
It's the same here, _Lietuvos_ (adjective) means "relating to the Lithuanian language or Lithuania", but language name is **Lietuvių** (noun).
- (Turkish) _Turkish_ ➟ **Türkçe**.
Use native name instead of an English one.

Links:
- [List of language names](https://en.wikipedia.org/wiki/List_of_language_names)


sikmir:
@jef-n Jürgen, could you please take a look at this PR?

PR #551, opened 2018-07-16
Update internal links in 3.2 changelog introduction

DelazJ:
Browsing the 3.2 changelog in a language other than English, the links in the intro part bring you to the English page instead of keeping the language being used. This is due to the untranslated hard link used.
This PR instead uses the rst internal references, ensuring that the language is always respected (regardless translation level) and the link will not be broken if ever the repository structure changes (files renamed, (re)moved...).


rduivenvoorde:
@DelazJ Thanks for fixing! Will try to think about it next time. Was in a hurry this time to do it fast...

PR #596, opened 2019-02-18
update roadmap for QGIS 3.4

Gustry:
I just wanted to know the lifetime of QGIS 3.4 approximately.

According to the history, 2.8 and 2.14 were LTR, so 2 two versions (2.10 and 2.12 were not). So 3.10 should be the next one, right?

@jef-n ?


DelazJ:
Yes, according to the current process ([three releases per year](https://qgis.org/en/site/getinvolved/development/roadmap.html#release-schedule)), 3.10 would be the next LTR.
I also understood that the schedule should not be manually updated as it's managed by https://github.com/qgis/QGIS-Website/blob/master/scripts/update-schedule.py

jef-n:
> I also understood that the schedule should not be manually updated as it's managed by https://github.com/qgis/QGIS-Website/blob/master/scripts/update-schedule.py

correct.

Gustry:
Oh ok, nice, I didn't know this spreadsheet and python script.
I hope someone with write permission will update the google document. (for the 3.4 lifetime)

Thanks

PR #609, opened 2019-03-19
Use internal reference instead of link to github

DelazJ:


DelazJ:
If ever someone knows why the check is still failing with this (an area the PR does not touch)...
> Warning, treated as error:
/home/travis/build/qgis/QGIS-Website/source/site/getinvolved/governance/finance/index.rst:9:download file not readable: /home/travis/build/qgis/QGIS-Website/source/static/site/getinvolved/governance/finance/annual-financial-reports/PublicQGISfinancialreport2018.pdf
make: *** [html] Error 1
The command "make html" exited with 2.

PR #610, opened 2019-03-19
Fix link and build

DelazJ:


DelazJ:
@andreasneumann A proper fix would be to rename the financial report file name like the others (no space) but i think it's safer/trustful that you are the one who does it (since it requires uploading file).

DelazJ:
supersed by fa7da0d and efc311de1

PR #620, opened 2019-03-27
Drop boundless from commercial support organisations

nyalldawson:
As far as I'm aware, Boundless no longer exist, and are no longer available for QGIS support.


nyalldawson:
@volaya?

PR #621, opened 2019-03-27
Drop DMS from support providers

nyalldawson:
DMS hasn't existed for a long time now


nyalldawson:
@NathanW2 ok by you?

NathanW2:
Yes.

PR #629, opened 2019-04-21
add japanese books

waigania13:
Could you add some Japanese books?

### DIFF
--- a/source/site/forusers/books/index.rst
+++ b/source/site/forusers/books/index.rst
@@ -264,3 +264,28 @@ The book consists of 2 main parts; In the first part, Geographical Information S
 * Authors: Can Ayday, Levent Sabah, Nefise Yaman, Kübra Keser
 * Date: January 2016
 
+Japanese
+======= 
+
+プロ直伝 伝わるデータ・ビジュアル術 ――Excelだけでは作れないデータ可視化レシピ
+*****************************************************************************
+
+* https://gihyo.jp/book/2019/978-4-297-10582-2
+* Authors: Kosuke ASAHI and Others
+* Date: December 2019
+
+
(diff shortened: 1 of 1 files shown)


rduivenvoorde (commented):
Hi, can you please do some changes (or let me know, I'm happy to fix them for you):
- in rst the number of characters below a title should have exact the same number as the characters, so
below Japanese you nee 8 '=' characters (Sphinx build will fail on this).

- below the Booktitle there has to be a line of  stars (*) too (but I can fix that by viewing at it later)

- going into google translation the last two book look really like QGIS books, but are you sure the first one is too ( there is even an arcgis chapter? ;-) )? We like to have 'QGIS books here' not so much 'General GIS' books, so maybe skip the first one? Or let me know when I'm wrong.

Regards,

Richard Duivenvoorde
ps happy to fix this for you if it appears to be not simple for you to do

waigania13:
I'm very happy if you fix the following two.

> * in rst the number of characters below a title should have exact the same number as the characters, so
>   below Japanese you nee 8 '=' characters (Sphinx build will fail on this).
> * below the Booktitle there has to be a line of  stars (*) too (but I can fix that by viewing at it later)

As you have pointed out, first one is 'General GIS' book. Please skip first one.

> * going into google translation the last two book look really like QGIS books, but are you sure the first one is too ( there is even an arcgis chapter? ;-) )? We like to have 'QGIS books here' not so much 'General GIS' books, so maybe skip the first one? Or let me know when I'm wrong.

rduivenvoorde:
@waigania13 Ok, thanks, done by hand, will close this PR. 
Will show up with new build
If you want you can sent me one or two lines as 'introduction' text, just as with most of the other books, happy to add that too.

PR #671, opened 2019-09-09
Add Chartis Technology to commercial support page

NathanW2:
Adds Chartis Technology as a Core Contributors list in the commercial support page.

(I work for Chartis Technology)


NathanW2:
Thanks

PR #674, opened 2019-09-11
add qtibia to commercial support

tudorbarascu:


source/site/forusers/commercial_support.rst
```diff
@@ -229,6 +229,15 @@ commercial support, custom programming and training for QGIS, PostGIS and GRASS.
 
 ----
 
+.. |qtibia| image:: /static/site/about/images/qtibia.png
+   :height: 200 px
+
+|qtibia| `QTIBIA Engineering <https://qtibia.ro//>`_ (based in Romania, operating worldwide)
+provides services, support and training on QGIS, Postgis and other FOSS.
+We are extra specialised in the water/wastewater sector.
```
andreasneumann: My suggestion is to change "We are extra specialised in" ... with "Our focus is on " ...

Other than that I would think we can merge this and I also think that the "Contributor" section is the correct section to use.

tudorbarascu: Thanks for the valuable feedback, I think it's good now.

Issue #608, opened 2019-03-18
Recommend alternative Mac installs


systemed:
Currently the download page links to a QGIS installer for ~~Mac OS X~~ macOS with this README:

![Screenshot 2019-03-18 at 13 41 09](https://user-images.githubusercontent.com/694425/54535264-13e91b00-4986-11e9-88fd-41d6b4948947.png)

This is contrary to current best practice on the Mac. Most developers use the [Homebrew package manager](https://brew.sh) and will often already have Python 3 and GDAL installed, which this installer is unable to pick up. An "unmanaged" installation in system directories is less easily maintainable and risks breaking existing installations. For non-developers, requiring separate installation of dependencies is unnecessarily complex.

Two alternatives exist:

* For developers, QGIS can be installed directly from Homebrew with `brew tap osgeo/osgeo4mac ; brew install qgis`, which will use existing GDAL/Python/etc. where possible;
* For developers and non-developers alike, @lutraconsulting provide one-click installers with all dependencies bundled directly in the package rather than in system locations: https://lutraconsulting.github.io/qgis-mac-packager/

I'd recommend linking to the Lutra Consulting installer as a default and mentioning Homebrew as an alternative.

M-Rick:
Thank you for this alternative built.
I tried it out and it even seems to work much better than the official built, especially with Spatialite.
qgis/QGIS#29714

twenty-firstcenturymania:
I am having a similar problem with the new QGIS update. the older 2.7 version is compatible with the python plugin but the new version is not recognizing python.

DelazJ:
@twenty-firstcenturymania please check whether the QGIS version you install is compatible with your OS version. There are two installers.
> I'd recommend linking to the Lutra Consulting installer as a default and mentioning Homebrew as an alternative.

@systemed is your report still valid? The download page has been updated in the meantime (the target is non-developer users so only the "lutraconsulting package" has to be mentioned there imho)

systemed:
The download page is great now.

PR #686, opened 2019-10-14
Replace MacOS X label with macOS

DelazJ:
refs #608


agiudiceandrea:
@DelazJ Should the sentence "QGIS is freely available on Windows, Linux, **MacOS X**, BSD, and Android." in [All Download](https://www.qgis.org/en/site/forusers/alldownloads.html) page be updated, too?

DelazJ:
@agiudiceandrea yes I think it should. Sorry I missed your message.

PR #689, opened 2019-10-22
Fix left margin on carousel caption links

agiudiceandrea:
The links in the captions of the home page banner have an unnecessary and unaesthetic 10 px left margin:
image
image
image

This patch should fix it, removing the .carousel-caption .caption-link left margin in qgis-style.less and qgis-style.css


rduivenvoorde:
Thanks for the fix!

PR #696, opened 2019-10-28
Fix two "and" conjunctions not translatable in download page

agiudiceandrea:
Fixes two "and" conjunctions that are not available for translation in the sentences: "More older releases are available here **and** for OS X here." and "QGIS Source Code is available here (latest release) **and** here (long term release)" in the [download](https://www.qgis.org/en/site/forusers/download.html) page,

Please, check if the fix actually fixes the bug before merging,

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -230,7 +230,8 @@
           <div class="tab-pane" id="tab2">
               <p>{{ _('Previous releases of QGIS are still available') }} <a href="https://qgis.org/downloads">{{ _('here') }}</a> {{ _(' - including older releases for OS X ') }}
               <a href="https://qgis.org/downloads/macOS/">{{ _('here') }}</a>.</p>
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
You are right, these two 'and' strings were not flagged as translatable.
Your commit fixes this.
Normally you would not see it, because those new 'strings' would have been pushed to transifex as new source. But I see/think dat 'and' already is a 'source-string' so should show up tomorrow.
I will also push new strings for the website this week anyway.
Thanks.

agiudiceandrea:
Thanks @rduivenvoorde. 

> I will also push new strings for the website this week anyway.

Could you please have a look at these suggestions: https://github.com/qgis/QGIS-Website/commit/a86e9f8297c76aa04cf625bc2826f7da76a7ca2a#r35767731 and https://github.com/qgis/QGIS-Website/pull/686#issuecomment-547189465 and eventually implement them before pushing the strings?

Thank you again.

PR #701, opened 2019-11-04
Updated the SUSE / openSUSE

badarotti:
Repos older than 15.0 were removed. Also added other architectures available in the repos.


rduivenvoorde:
@dassau can you confirm (and pull) ?
@badarotti I'm totally not in the Suse infra, so cannot check this. Will pull anyway if nobody responds (next build of site will be tomorrow anyway)

PR #725, opened 2020-03-03
updated links to Input ios and ms windows and also sorted them alphab…

saberraz:
…etically

See:
https://github.com/qgis/QGIS-Website/pull/724#issuecomment-594152333

PR #727, opened 2020-03-04
added ios link to input due to premature closure of my original PR

saberraz:
Follow up to https://github.com/qgis/QGIS-Website/pull/725


timlinux:
Ah cool! Thanks for this @saberraz, I will merge it!

saberraz:
Now we have apps for all platforms....yayyy
Next target browsers and smart watches!

vpicavet:
Hello,
I am worried that this could lead to legal troubles. There are clear issues with providing GPL applications on apple store. While having an app supporting iOS could be ok, distribution through the Apple app store makes the user have to accept iOS terms of use, which are not compatible with the licence of the software.

While publishing GPL software on iOS is the issue of the software provider / publisher, linking it from qgis.org website would be an endorsement, and would lead to legal risk. 

I would personally be against it, and at least it should be an official PSC decision, having taken into account all related risks, preferably after seeking official legal advice from an opensource IP expert lawyer.

I copy hereafter the report of our IP lawyer, who we consulted when we had this question a while ago (on a different GPL software) : 

## Concerning publishing a GNU GPL application on App Store

Distributing software on Apple App Store is likely to contrain the user
to accept General Usage Conditions s a-priori not compatible with the
GNU GPL Licence v3.0 or later. As a matter of fact, AppStore TOS --
which users as well as developers must accept -- constrain the usage
that can be done with the software, which is contradictory to the GNU
GPL v3 or later. The latter forbids to apply such usage restrictions
when distributing a software under its terms.

Such a difficulty also applies to the distribution of a version of
software X on AppStore including Z ( licenced as GPL ). It is on this
basis that the application VLC has been initially removed from Appstore,
before being re-introduced under a new licence conform ( double licence
Mozilla Public Licence 2.0 and GNU GPL v3.0).

Therefore, such a distribution would be possible under appstore only if
its licence is modified according to this choice.

Article 10 of GNU GPL 3.0 : «  You may not impose any further
restrictions on the exercise of the rights granted or affirmed under
this License. For example, you may not impose a license fee, royalty, or
other charge for exercise of rights granted under this License, and you
may not initiate litigation (including a cross-claim or counterclaim in
a lawsuit) alleging that any patent claim is infringed by making, using,
selling, offering for sale, or importing the Program or any portion of it. »

pcav:
Thanks Vincent. @saberraz do you have evidence this is not the case? Obviously QGIS.org cannot endorse licence violations.

saberraz:
Feel free to revert the PR. But I am still failing to see the risk and these types of arguments just make life more difficult for end users who live in real world.

vpicavet:
Well, if we want our users to respect the GPL licence which is the basis for all QGIS work, and share their modifications back, we cannot compromise on legal aspects, even if it is a burden for the user.
This is sad, but "real world people do not respect licences" is not really an argument.

PR #736, opened 2020-03-19
remove direct link to QGIS for android.

mbernasocchi:
As per PSC decision since this is which is deprecated
image


mbernasocchi:
thanks @rduivenvoorde

PR #737, opened 2020-03-20
Fix index for testing documentation

agiudiceandrea:
QGIS testing now is "newer than QGIS ~~3.4~~ 3.10" and "We call this version 'QGIS testing' and _the documentation_ can be found here:"


DelazJ:
Thanks @agiudiceandrea

PR #746, opened 2020-04-07
renew OGC badge certicate for WMS 1.3

haubourg:
Hi there, 
I finally managed to renew the OGC certification for QGIS server. Here is the updated badge.


pblottiere:
> Here is the updated badge.

Nice, thanks Régis  :+1:

PR #769, opened 2020-05-25
fix the spell error of chinese

sshuair:
The translate spell of Chinese (Simplified) and Chinese (Traditional) is not correct.

- `簡体中文` to `简体中文`
- `正體中文` to `繁體中文`

### DIFF
--- a/themes/qgis-theme/languageswitch.html
+++ b/themes/qgis-theme/languageswitch.html
@@ -25,8 +25,8 @@
   <option value="tr" title="{{ _('Turkish') }}">Türkçe</option>
   <option value="ru" title="{{ _('Russian') }}">Русский</option>
   <option value="uk" title="{{ _('Ukrainian') }}">Українська</option>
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
@sshuair can you please explain the differences, just out of curiousity.
And I think we have this on the documentation for 3.10 then too isn't it?

sshuair:
- Chinese (Simplified)
`簡` is Traditional Chinese characters  and `简` is Simplified Chinese characters, so `Chinese (Simplified)` should be `简体中文` not `簡体中文`.

- Chinese (Traditional)
`正體中文` is used more in Taiwan, but `繁體中文` is used in Taiwan, Hong Kong, Singapore and Macau, so `繁體中文` is more widely used.

Another reason is QGIS desktop is `繁體中文`. MacOS and Windows is also used `繁體中文`.

image

rduivenvoorde:
Ok, thanks. I also redirect zh_CN to zh-Hans now and zh_TW to zh_Hant. 
I will have to rebuild all to make it better, but please keep checking and reporting. This is messy...

PR #771, opened 2020-05-25
replace screenshot of qgis on android with mobile apps

mbernasocchi:
@saberraz FYI for input I took the first 3 images from the playstore (same as for QField), @NathanW2  I took only one picture since all I could found is in landscape mode.


saberraz:
that is great @mbernasocchi 

Let me use a more recent screenshots of Input.

PR #777, opened 2020-06-04
Easy debian/ubuntu instructions

rduivenvoorde:
After some discussion on IRC: http://irclogs.geoapt.com/qgis/%23qgis.2020-06-04.log

Keesj created a gist: https://gist.github.com/keesj/794e038f3efe0e0b383fce1a1b16c110

I made it even shorter so it will only work for latest stable.

What do you think? Please also test.

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -102,7 +102,7 @@ versions of QGIS in different versions of Ubuntu.
 Here you will simply install the latest stable QGIS (|version|.x |codename|)
 in your Ubuntu or Debian without having to edit config files.
 
-Note: although you see 'Ubuntu' in some places, this also works for 'Debian'.
+.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian'.
 
 First install some tools you will need for this instructions::
 
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -108,7 +108,7 @@ First install some tools you will need for this instructions::
 
  sudo apt-get install gnupg software-properties-common
 
-Now we install the QGIS Signing Key, so QGIS software from 
+Now install the QGIS Signing Key, so QGIS software from 
 the QGIS repo will be trusted and installed::
 
  sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 51F523511C7028C3
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -102,7 +102,7 @@ versions of QGIS in different versions of Ubuntu.
 Here you will simply install the latest stable QGIS (|version|.x |codename|)
 in your Ubuntu or Debian without having to edit config files.
 
-.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian'.
+.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian', as one is actually a symlink to the other on our server.
 
 First install some tools you will need for this instructions::
 
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -93,10 +93,10 @@ there are instructions to install QGIS stable and instructions to install a
 cutting edge QGIS testing build (note the warning_).
 
 
-Debian/Ubuntu for beginners
----------------------------
+Debian/Ubuntu
+-------------
 
-In the section following this one, you will find ALL possible options to install different
+In the 'advanced' section following this one, you will find ALL possible options to install different
 versions of QGIS in different versions of Ubuntu.
 
 Here you will simply install the latest stable QGIS (|version|.x |codename|)
@@ -130,8 +130,8 @@ Note: add 'qgis-server' to this line if you also want to install QGIS Server::
  sudo apt-get install qgis qgis-plugin-grass
 
 
-Debian/Ubuntu
--------------
+Debian/Ubuntu (advanced)
+------------------------
 
 Default Debian and Ubuntu software repositories often hold older versions of
 QGIS.
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -93,6 +93,43 @@ there are instructions to install QGIS stable and instructions to install a
 cutting edge QGIS testing build (note the warning_).
 
 
+Debian/Ubuntu for beginners
+---------------------------
+
+In the section following this one, you will find ALL possible options to install different
+versions of QGIS in different versions of Ubuntu.
+
+Here you will simply install the latest stable QGIS (|version|.x |codename|)
+in your Ubuntu or Debian without having to edit config files.
+
+Note: although you see 'Ubuntu' in some places, this also works for 'Debian'.
+
+First install some tools you will need for this instructions::
+
+ sudo apt-get install gnupg software-properties-common
+
+Now we install the QGIS Signing Key, so QGIS software from 
+the QGIS repo will be trusted and installed::
+
+ sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 51F523511C7028C3
+
+Add the QGIS repo for the latest stable QGIS (|version|.x |codename|).
+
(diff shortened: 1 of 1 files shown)


source/site/forusers/alldownloads.rst
```diff
@@ -93,6 +93,43 @@ there are instructions to install QGIS stable and instructions to install a
...
+Debian/Ubuntu for beginners
+---------------------------
+
+In the section following this one, you will find ALL possible options to install different
+versions of QGIS in different versions of Ubuntu.
+
+Here you will simply install the latest stable QGIS (|version|.x |codename|)
+in your Ubuntu or Debian without having to edit config files.
+
+Note: although you see 'Ubuntu' in some places, this also works for 'Debian'.
```
DelazJ:
```suggestion
.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian'.
```

source/site/forusers/alldownloads.rst
```diff
@@ -93,6 +93,43 @@ there are instructions to install QGIS stable and instructions to install a
...
+
+First install some tools you will need for this instructions::
+
+ sudo apt-get install gnupg software-properties-common
```
DelazJ, line 109:
```suggestion
 sudo apt install gnupg software-properties-common
```
isn't it? And below...
I wanted to suggest to drop the sudo in codes and advise the context of use in the note above but i see you discussed it already in irc... :thinking:

source/site/forusers/alldownloads.rst
```diff
@@ -93,6 +93,43 @@ there are instructions to install QGIS stable and instructions to install a
...
+
+Now we install the QGIS Signing Key, so QGIS software from
```
DelazJ:
```suggestion
Now install the QGIS Signing Key, so QGIS software from 
```

source/site/forusers/alldownloads.rst
```diff
@@ -96,6 +96,43 @@ cutting edge QGIS testing build (note the warning_).
...
 
+In the 'advanced' section following this one, you will find ALL possible options to install different
+versions of QGIS in different versions of Ubuntu.
+
+Here you will simply install the latest stable QGIS (|version|.x |codename|)
+in your Ubuntu or Debian without having to edit config files.
+
+Note: although you see 'Ubuntu' in some places, this also works for 'Debian'.
+
+First install some tools you will need for this instructions::
```
DelazJ, line 107: Could we go with numbered list here, showing things as real steps?
And should it be "this" or "these"?

source/site/forusers/alldownloads.rst
```diff
@@ -96,6 +96,43 @@ cutting edge QGIS testing build (note the warning_).
...
+ sudo apt-get install gnupg software-properties-common
+
+Now we install the QGIS Signing Key, so QGIS software from 
+the QGIS repo will be trusted and installed::
+
+ sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 51F523511C7028C3
+
+Add the QGIS repo for the latest stable QGIS (|version|.x |codename|).
+
+Note: "lsb_release -c -s" in those lines will return your distro name ::
```
DelazJ:
```suggestion
Note: "lsb_release -c -s" in those lines will return your distro name::
```

DelazJ: Actually I'd have seen this as a `.. note::` after the snippet with adding mention to replace https://qgis.org/ubuntu with corresponding url at (put link to the advanced table here) if they want another version.

Or it becomes less beginner-target? Though being a beginner in Ubuntu does not mean that you are a beginner in QGIS use/testing (I know what i'm talking about :smile: ) so you may want something else than latest stable...

DelazJ (commented):
Thanks to you, guys.
I think this part of the installation process is the one that has got the greatest number of attempts of rewords and questions in ML, revealing that we are missing something. Splitting and having a beginner section will help many people find their way in the QGIS installation process IMO.

source/site/forusers/alldownloads.rst
rduivenvoorde: apt is a script around apt-get, but both are still available. I would leave it as is.

about sudo: this is the easiest way to make it copy/pastable for ubuntu users. So I prefer to keep it like this.

source/site/forusers/alldownloads.rst
rduivenvoorde: Don't know.

source/site/forusers/alldownloads.rst
```diff
@@ -96,6 +96,43 @@ cutting edge QGIS testing build (note the warning_).
 Debian/Ubuntu
 -------------
 
+In the 'advanced' section following this one, you will find ALL possible options to install different
+versions of QGIS in different versions of Ubuntu.
+
+Here you will simply install the latest stable QGIS (|version|.x |codename|)
+in your Ubuntu or Debian without having to edit config files.
+
+.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian'.
```
keesj:
```suggestion
.. note:: The documentation here is documenting the ubuntu flow but the procedure has also been tested on debian.  When following these instructions for debian replace https://qgis.org/ubuntu by https://qgis.org/debian in the instruction bellow.
```

rduivenvoorde: I wanted to keep it as small as possible. And (I tested) you do not need to replace ubuntu by debian, as both url's point to the same webdir in the repo, and there the debian and ubuntu versions are mixed: 

https://qgis.org/ubuntu/dists/

https://qgis.org/debian/dists/

So it 'will just work'? Or do I miss some info here?

keesj: I do not know the rationale behind the choice of having different repositories. If they are supposed to be the same and will remain the same it will just work. e.g. we have only tried it correct, not proved it. I personally would be a bit suspicious if I was running debian and was using an ubuntu repository. can we check with the repository maintainer?

jef-n: There's is just one repository.   debian gets the packages,  ubuntu is just a symlink.   Introduced for those users who apparently didn't get that they could have used the /debian/ repo for ubuntu and used the /ubuntugis/ repo instead, just because that has ubuntu in it's name.

rduivenvoorde: So are we ok with keeping 'ubuntu' in the repo url's, because I think Debian users will read the note and be fine?
I just want to keep this as small as possible, adding more details is watering down the message I think. IF we ever split up the repo, off course we have to change this.

keesj: Yes . but I think we need to tell that the directories are are symlink

keesj: .. note:: Although the documentation appears to be targeting Ubuntu this also works on Debian without modification. If running on debian feel free to replace https://qgis.org/ubuntu by https://qgis.org/debian in the instruction bellow.

rduivenvoorde: Done

DelazJ: I don't think it's necessary. These are internal information, and whatever they use, it will work so mentioning it wouldn't add anything imho. Let's not forget that we are in a beginner-targeted section. (and i agree with @rduivenvoorde that probably people will switch to the OS they use and if they don't there's no issue)

DelazJ: looks like my page did not refresh before i sent my message. so too late. And I'm good with the rewording

source/site/forusers/alldownloads.rst
```diff
@@ -102,7 +102,7 @@ versions of QGIS in different versions of Ubuntu.
 Here you will simply install the latest stable QGIS (|version|.x |codename|)
 in your Ubuntu or Debian without having to edit config files.
 
-.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian'.
+.. note:: Although you see 'Ubuntu' in some places, this also works for 'Debian', as one is actually a symlink to the other on our server.
```
keesj, line 105: .. note:: Although you see ubuntu in the repository name this also works for Debian, as one is actually a symlink to the other on our server.

PR #778, opened 2020-06-04
Fix Sample Dataset download link

agiudiceandrea:
In the For Users / Download QGIS page at https://qgis.org/en/site/forusers/download.html, the link to download a sample dataset points to a nonexistent http://docs.qgis.org/3.12/en/docs/user_manual/introduction/getting_started.html#sample-data
as `http://docs.qgis.org/3.12/` is nonexistent and the "Downloading sample data" paragraph of the "Getting Started" chapter is at `/getting_started.html#downloading-sample-data`.

To fix this, I propose to change the link to `http://docs.qgis.org/latest/{{language}}/docs/user_manual/introduction/getting_started.html#downloading-sample-data`

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -236,12 +236,12 @@
               </div>
               <!-- End Accordion -->
 
-              <h4><a href="{{ pathto('site/forusers/alldownloads') }}">{{ _('All downloads') }}</a></h4>
+              <h4>{{ _('All downloads') }}</h4>
               <p>
                 {{ _('More specific instructions about downloading QGIS stable vs QGIS development can be found in') }} <a href="{{ pathto('site/forusers/alldownloads') }}">{{ _('All downloads') }}</a>.
               </p>
 
-              <h4><a href="http://docs.qgis.org/latest/{{language}}/docs/user_manual/introduction/getting_started.html#downloading-sample-data">{{ _('Datasets') }}</a></h4>
+              <h4>{{ _('Datasets') }}</h4>
               <p>{{ _('For testing and learning purposes, a ') }} <a href="http://docs.qgis.org/latest/{{language}}/docs/user_manual/introduction/getting_started.html#downloading-sample-data"><strong>{{ _('sample dataset is available') }}</strong></a>, {{ _('which contains collections of data from different sources and in different formats.') }}</p>
 
           </div>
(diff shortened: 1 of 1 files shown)


DelazJ:
Hmmm... Actually there are two issues:
- a redirection of 3.12 that was not working as expected: everything >3.10 should point to the same url, but in master, and something was broken in the redirection. Fixed now, and it should point to the "getting started" file of testing doc. It does however not point to the actual section because of :arrow_down: 
- the section title that has changed; fixed in the PR

About moving from version to latest, it's not that a bad idea given that in most of the case, it will keep serving the instructions in the user language (which is a good experience) instead of turning it into the master English.

themes/qgis-theme/forusers_download.html
```diff
@@ -241,8 +241,8 @@ <h4><a href="{{ pathto('site/forusers/alldownloads') }}">{{ _('All downloads') }
                 {{ _('More specific instructions about downloading QGIS stable vs QGIS development can be found in') }} <a href="{{ pathto('site/forusers/alldownloads') }}">{{ _('All downloads') }}</a>.
               </p>
 
-              <h4><a href="http://docs.qgis.org/{{version}}/{{language}}/docs/user_manual/introduction/getting_started.html">{{ _('Datasets') }}</a></h4>
-              <p>{{ _('For testing and learning purposes, a ') }} <a href="http://docs.qgis.org/{{version}}/{{language}}/docs/user_manual/introduction/getting_started.html#sample-data"><strong>{{ _('sample dataset is available') }}</strong></a>, {{ _('which contains collections of data from different sources and in different formats.') }}</p>
+              <h4><a href="http://docs.qgis.org/latest/{{language}}/docs/user_manual/introduction/getting_started.html#downloading-sample-data">{{ _('Datasets') }}</a></h4>
```
DelazJ: I was wondering (never noticed until now): Could this just not be a section title? Do we need to make it a U!RL here, considering that the same url is provided in the next sentence?

agiudiceandrea: I thought so too, but also the previous section title has a link which is the same as the one in the section text:

> [All downloads](https://qgis.org/en/site/forusers/alldownloads.html)
> More specific instructions about downloading QGIS stable vs QGIS development can be found in [All downloads](https://qgis.org/en/site/forusers/alldownloads.html).

Should both the sections title link be removed?

agiudiceandrea:
Please, see also https://github.com/qgis/QGIS-Documentation/pull/5619

themes/qgis-theme/forusers_download.html
DelazJ: > Should both the sections title link be removed?

I'd say yes, no need to have a url in the sections title.

agiudiceandrea: Links removed from sections title.

DelazJ:
Thanks Andrea

PR #779, opened 2020-06-05
Update qugsnews.atom

andreasneumann:


rduivenvoorde:
Thanks Andreas, Raymond asked me to not remove 's Hertogenbosch yet, so I will do some refrasing afterwards...

PR #782, opened 2020-06-17
fix date of QGIS North America event

kgjenkins:


rduivenvoorde:
Thanks for the fix, and sorry for this mistake...

Issue #783, opened 2020-06-21
Broken link to slackware packages


openstrike:
## Description

The link to download Slackware packages goes to http://qgis.gotslack.org/ but that FQDN does not exist. The link should ideally be updated to wherever the Slackware packages currently reside.

Page URL: https://www.qgis.org/en/site/forusers/alldownloads.html#slackware

rduivenvoorde:
@openstrike not being an Slackware user myself I tried to google where they have packages.
One thing I see is searching here for 'qgis':
https://packages.slackware.com/
does not reveal packages...

Most 'news' links are 2016 or 2017, is Slackware still alive? Or should we just remove it from the Distro list untill somebody comes with the right place?

openstrike:
Yes, Slackware is very much still alive (see eg. the [current changelogs](http://www.slackware.com/changelog/current.php?cpu=x86_64)). We are using it which is why I was looking at the link in the first place.

The only semi-official reference to qgis I could find is the slackbuild entries such as https://slackbuilds.org/repository/14.2/gis/qgis/?search=qgis so perhaps that's the most appropriate link for now? It is probably what I would use to install it.

rduivenvoorde:
@openstrike are you able to do a PR for this?

openstrike:
Sure - leave it with me.

DelazJ:
Fixed in #784

PR #784, opened 2020-06-23
Update slackware URL

openstrike:
This PR should replace all instances of the obsolete http://qgis.gotslack.org URL as per issue #783. 

Note that I have removed the trailing "-on" from the Hungarian translation as that would appear to interfere with the URL.

### DIFF
--- a/i18n/ca/LC_MESSAGES/site/forusers/alldownloads.po
+++ b/i18n/ca/LC_MESSAGES/site/forusers/alldownloads.po
@@ -672,8 +672,8 @@ msgid "Slackware"
 msgstr "Slackware"
 
 #: ../../source/site/forusers/alldownloads.rst:372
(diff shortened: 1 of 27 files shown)


rduivenvoorde (changes requested):
@openstrike Thanks! 
Actually you only have to do the one line change in alldownloads.rst
Translation is done via transifex.com and their version is the latest, so what is in git (with us) is actually an old version...
IF we change something (and push to tx) then the 'msg'-id will change and all other languages are build with the english string (untill the translator translates it).

In short: please revert all the po file changes (or drop this pr and create an inimini one with only the rst change)

Thanks!

openstrike:
The .po files have been reverted now.

rduivenvoorde:
Thanks!

PR #786, opened 2020-06-23
Update devversion and nextversion in schedule.py

agiudiceandrea:
Update devversion to '**3.15**' and nextversion to '**3.16**' in schedule.py.
They are incorrectly reported, respectively, as **3.13** and **3.14** in version checking information:

> The current released version of QGIS is 3.14.0.
> We are working on **3.13**-Master. It will go into feature freeze on 2020-09-11 12:00:00 UTC and be released as **3.14** on 2020-10-23 12:00:00 UTC.

and in https://qgis.org/it/site/forusers/alldownloads.html

> Development | **3.13** master | qgis-dev

> Development Version | **3.13** master | Nightly build for Debian and Ubuntu


agiudiceandrea:
fixed in 7519d17

PR #810, opened 2020-09-28
[Get Involved][FAQ] Update types of citations

jonnyforestGIS:
Update this section:
https://qgis.org/en/site/getinvolved/faq/index.html#how-to-cite-qgis

Tasks:

- [x] - Copy information from https://issues.qgis.org/projects/qgis/wiki/QGIS_Citation_Repository
- [x] - Update links from citations

If it is necessary to change something, please let me know.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -81,7 +81,7 @@ exec(open("schedule.py").read())
 # non-false value, then it is used:
 # today = ''
 # Else, today_fmt is used as the format for a strftime call.
-# today_fmt = '%B %d, %Y'
+today_fmt = '%Y'
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/source/conf.py
+++ b/source/conf.py
@@ -223,7 +223,23 @@ rst_epilog = docs_rst_epilog + """
 .. |checkbox_unchecked| image:: /static/site/common/checkbox_unchecked.png
 .. |QG| replace:: QGIS
 .. |qg| replace:: QGIS
-"""
+.. |devcite| raw:: html
+
+     <a href="https://docs.qgis.org/%s/en/docs/developers_guide/index.html"> https://docs.qgis.org/%s/en/docs/developers_guide/index.html</a>
+
+.. |userguidecite| raw:: html
+
+     <a href="https://docs.qgis.org/%s/en/docs/user_manual/index.html"> https://docs.qgis.org/%s/en/docs/user_manual/index.html</a>
+
+.. |servercite| raw:: html
+
+     <a href="https://docs.qgis.org/%s/en/docs/server_manual/index.html"> https://docs.qgis.org/%s/en/docs/server_manual/index.html</a>
+
+.. |apicite| raw:: html
+
+     <a href="https://qgis.org/pyqgis/%s/index.html"> https://qgis.org/pyqgis/%s/index.html</a>
+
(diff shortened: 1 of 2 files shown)


rduivenvoorde (commented):
Thanks! 
One comment on the BitTex part: should that not be updated? Like: point the url to httpS://qgis.org (you are being redirected there anyway), and year: either make it 2020 or automagically 'this' year, Foundation: we can keep it like that, OR change it to the QGIS Association (https://www.qgis.org/en/site/getinvolved/governance/charter/index.html)?

And did we not have a Server Manual now too (https://docs.qgis.org/testing/en/docs/server_manual/index.html)?

jonnyforestGIS:
Hi @rduivenvoorde , 

> One comment on the BitTex part: should that not be updated? 
>Like: point the url to httpS://qgis.org (you are being redirected there anyway)
You are right  I forgot to update this part.

> year: either make it 2020 or automagically 'this' year
This is related to substitutions in sphinx? I don't know how to do it but I can read about it.

> Foundation: we can keep it like that, OR change it to the QGIS Association 
>(https://www.qgis.org/en/site/getinvolved/governance/charter/index.html)`
I agree. I can change as well 

> And did we not have a Server Manual now too (https://docs.qgis.org/testing/en/docs/server_manual/index.html)?
Good catch Richard.

Thank you for your feedback @rduivenvoorde .

jonnyforestGIS:
@rduivenvoorde,

Tasks:
   
- [x]  One comment on the BitTex part: should that not be updated? Like: point the url to httpS://qgis.org (you are being redirected there anyway)
   
- [x] year: either make it 2020 or automagically 'this' year.

  **Note:** I achieve the substitution for the year and version in the text and inside of BibTeX.  I assume the |ltrversion| substitution. I need to search how can we can pass the |ltrversion| substitution inside of url.
 
- [x] Foundation: we can keep it like that, OR change it to the QGIS Association (https://www.qgis.org/en/site/getinvolved/governance/charter/index.html)`


- [x]   And did we not have a Server Manual now too (https://docs.qgis.org/testing/en/docs/server_manual/index.html)?

Please feel free to change you guide me to correct the missing task.

jonnyforestGIS:
I was able to do the substitution for the URLs assuming de variable ltrversion because is the long-term release. If necessary, can be changed for the latest version.

I think this is ready to merge.  Please let me know if something is missing.

jonnyforestGIS:
gentle ping @rduivenvoorde. Can you review these last changes and see if they're ok?  Cheers.

source/conf.py
```diff
@@ -81,7 +81,7 @@
 # non-false value, then it is used:
 # today = ''
 # Else, today_fmt is used as the format for a strftime call.
-# today_fmt = '%B %d, %Y'
```
rduivenvoorde: conf.py is the configuration file for ALL website code. So by changing this you would break all other places where a 'today_fmt' would be used. So unless you are 100% sure it is never used... do not do this.

Better to use something like this:
https://gitlab.idiap.ch/bob/bob.learn.em/blob/master/doc/conf.py
(put it as our own substitution in the prolog)

jonnyforestGIS:
Hi again @rduivenvoorde, sorry for the delay. I add your review and create a substitution for the year.

jonnyforestGIS:
Hi @rduivenvoorde gentle ping, do you need more changes in this or this can be merged?
Cheers

rduivenvoorde:
@jonnyforestGIS sorry. This is fine, let's pull! Thanks!!

PR #811, opened 2020-09-30
Add note about xdg-desktop-portal for Flatpak

divergentdave:
This PR will add a note to the Flatpak-based installation instructions about the xdg-desktop-portal/xdg-desktop-portal-gtk packages. Users on certain distributions will need to install these packages in addition to Flatpak itself, or else file dialog boxes won't open. See previous discusson on flathub/org.qgis.qgis#32.


rduivenvoorde:
Thanks for adding this note!

PR #814, opened 2020-10-12
No 3.16 visual changelog yet

DelazJ:
QGIS-Website is failing. @rduivenvoorde Is there a plan to pull the changelogs in upcoming hours or should we fix the build now?


rduivenvoorde:
Sorry accidently commited this file which I use to test current rst exports...
No I think we are trying to do it directly on day of release, so no need to point to the changelog application I think
Thanks for taking care.

PR #817, opened 2020-10-19
Added openSUSE 15.2 repository

badarotti:
Thanks for all the great work.


rduivenvoorde:
Thanks. At what rate does OpenSuse release, or actually can we maybe cleanup 15.0?

PR #829, opened 2020-10-26
Remove URLs from translatable strings

agiudiceandrea:


rduivenvoorde:
Yep, good catch. Thanks!

PR #835, opened 2020-11-02
Use GitHub action for an HTML build of the website

DelazJ:
probably faster...


rduivenvoorde:
@DelazJ does this work already? So this is then an alternative to Travis yes? To test if a PR breaks the build or not?

DelazJ:
Yes a faster alternative to travis. This is mainly a copy-paste of what we have for docs (test the html build and create a zip for the build - maybe should I remove that part?) so it should work.

PR #838, opened 2020-11-13
Update qugsnews.atom, add missing country

gacarrillor:


juliogontor:
thanks @gacarrillor i miss that

PR #849, opened 2020-12-26
Split the 3.16 blocks into groups as of 3.10 and docs website

DelazJ:
Also add links to download the html zip (and pdf) for the releases in a new list. Looks like this
image

@anitagraser are you the person taking care of the version number image shown above? In which case, could you update that please? I found a doc.xcf file in the repo but no layer with 3 digits I can just change values. Would be nice to have that in the file. Thanks

### DIFF
--- a/themes/qgis-theme/docs_index.html
+++ b/themes/qgis-theme/docs_index.html
@@ -58,17 +58,33 @@
             <div class="span1"></div>
             <div class="offset1 span5">
                 <h3>{{ _('QGIS 3.16') }}</h3>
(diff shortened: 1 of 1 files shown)


anitagraser:
Yes, I've uploaded new files in https://github.com/qgis/QGIS-Website/commit/899c6d453ab21bda004b685bc7a43350b9ced2fb?short_path=5424aa8#diff-5424aa80b5befafd850a70f1a941ab374a870a8ad522a3e9b73151438f508aa5

@rduivenvoorde sorry for the Christmas delay, Richard! Good to see the new docs coming.

DelazJ:
Thanks @anitagraser and sorry, I didn't know that Richard already made the request.

PR #855, opened 2021-01-06
Add Ecuador

anitagraser:


mbernasocchi:
Ah sorry I missed this and already dit it

PR #866, opened 2021-02-23
Update index.rst

TurboGraphxBeige:
One "Feature" has been split in two. The title was for one feature but described two features. Also the description of the features were a little bit off so I reworded them in a cleaner way.


rduivenvoorde:
@TurboGraphxBeige thanks! I updated the source of index.rst taking your fixes into account, so closing this one as it is not needed anymore. But Thanks!

PR #875, opened 2021-03-02
QGIS is GPLv2+

vpicavet:
We should make it clear in the contributor guidelines that QGIS is GPL version 2 or above, as stated in the project's README.

### DIFF
--- a/source/site/getinvolved/development/contributor_requirements.rst
+++ b/source/site/getinvolved/development/contributor_requirements.rst
@@ -13,7 +13,7 @@ project, all developers/team members must:
   in the Development Corner, and the CODING document in the root of the QGIS
   Sources.
 * Ensure that all code committed to the repository is unencumbered and compatible
-  with the QGIS license, GPL version 2 or above
+  with the QGIS license, GNU Public License (GPL) Version 2 or any later version.
 * Not use or include any code that will cause a trademark or patent infringement.
   If you are unsure, ask on the developer mailing list or contact a Project
   Steering Committee member.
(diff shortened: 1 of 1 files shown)


source/site/getinvolved/development/contributor_requirements.rst
```diff
@@ -13,7 +13,7 @@ project, all developers/team members must:
   in the Development Corner, and the CODING document in the root of the QGIS
   Sources.
 * Ensure that all code committed to the repository is unencumbered and compatible
-  with the QGIS license, GPL version 2
+  with the QGIS license, GPL version 2 or above
```
mach0: Hi. just stumbled across this and.. Isn't the term usually
GPL version 2 or later ?
That would sound more common to me.

vpicavet: I took the expression from the project's README. I agree that "or later" is more frequently used. I would just try to keep a coherency, and change the README if we use "or later" here too.

mach0: the most used term I could find everywhere was: GNU Public License (GPL) Version 2 or any later version.
and that sounds nice and is also now stated in the README of QGIS.
If you'd like to change that I am going to pull the last PR in ;)

agiudiceandrea:
```suggestion
  with the QGIS license, GNU Public License (GPL) Version 2 or any later version.
```

agiudiceandrea (changes requested):

PR #881, opened 2021-04-08
add QGIS Ghana

enockseth:


andreasneumann:
Thank you! And good luck with the establishment of the user group Ghana!

PR #891, opened 2021-05-16
add 2020 report

mbernasocchi:
the PDF to be added is at https://drive.google.com/file/d/1fwA2ys9fZkFel7xsa-uxoilpOmXfx8s0/view?usp=sharing
@rduivenvoorde how do I add the pdf to the server? or can you add it and let me know how you did? 
Thanks

### DIFF
--- a/source/site/getinvolved/governance/annual_reports/index.rst
+++ b/source/site/getinvolved/governance/annual_reports/index.rst
@@ -9,6 +9,7 @@ highlighting key activities within the QGIS project.
 Note that the report will contain some financial information which is also
 available as a separate report in the finances section of this web site.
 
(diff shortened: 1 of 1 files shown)


DelazJ:
@mbernasocchi Some notes. I think there's an issue in the QGIS releases section of the report: the first row of the table is also the last one.
> Marco Bernasocchi (vice-chair until 2020-06; chair from 2020-05)

It should probably be `vice-chair until 2020-05` to be coherent with Alessandro date and not having you both chair and vice-chair.

rduivenvoorde:
@mbernasocchi you should put it here:

https://github.com/qgis/QGIS-Website/tree/master/resources/en/site/getinvolved/governance/annual-project-reports

it is a resource and during build will be copied then to the languages...

If not let me know. Not copying it now because there is apparently something wrong acording to Delazj...

Let me know if you need more

rduivenvoorde:
@mbernasocchi is this to be merged? Did you add the pdf?

PR #895, opened 2021-05-28
Update chat-options in support page

rduivenvoorde:
Because I think we should abandon Freenode, and think that Matrix is the best alternative I added Matrix (and Telegram, my second best) to the chat part of the support page.

I hope it is fine to share the Telegram-join link? It is already public via the github wiki pages, but feel free to remove it if you think this will propagate to spam in the group...


jmckenna:
Looks good here.

anitagraser:
Looks good, Richard. Do you want me to merge it?

PR #899, opened 2021-06-10
Add youtube channel banner

anitagraser:


rduivenvoorde:
Let me merge first, then I will build locally... and fix if needed

rduivenvoorde:
Screenshot-20210610162904-1103x1252

@anitagraser Not very pretty, do you think?

rduivenvoorde:
Not sure why span4 moved it to another row, that should be ok...
I'll make it span3 for now then it looks like this:

Screenshot-20210610163318-1410x402

anitagraser:
Weird ... and annoying. I paid extra attention to get it align nicely and with span3 everything looks wonky.

anitagraser:
There's an extra div class row in the case where 3x span4 works:

https://github.com/qgis/QGIS-Website/blob/89d43074ce7a8346ea92a5bdc6853633fad89d03/themes/qgis-theme/index.html#L118-L124

anitagraser:
Alternatively, I can rescale the logos, since I noticed none of them are shown in 100% anyway

anitagraser:
The row div seems to work:

image

rduivenvoorde:
Do what you think is right, I'm ok with both solutions.

PR #904, opened 2021-06-19
Arch Linux spelling fix

svetlemodry:
The correct spelling is "Arch Linux".
Official convention noted down in the Arch wiki [here](https://wiki.archlinux.org/title/Arch_terminology#Arch_Linux).


rduivenvoorde:
hi @svetlemodry, thanks for the fix. But we cannot 'feed back' the fixes in the .po (translation) files, because we 'pull' them from transifex only (and only push english source files (rst and html))

If you remove them from this PR and only fix the rst and html files, then with the next build all languages will probably ok because there is also a new 'translation string'...
(We could maybe also merge your po file fixes, but they will be overidden next build when we pull the translations from transifex...)

Ok?

rduivenvoorde:
@svetlemodry see the build fail message:

 Warning, treated as error:
/home/runner/work/QGIS-Website/QGIS-Website/source/site/forusers/alldownloads.rst:437:Title underline too short.

In rst the 'underline' which denotes a chapter heading should be exactly the amount of characters as the string it is underlining.
Can you please fix? Then the build check will run automagically.

svetlemodry:
And I thought this will be a boring pull request :)
Thanks for the quick replies and follow up @rduivenvoorde

rduivenvoorde:
Argh rst is a b*tch sometimes... Tables are also very picky. I'll merge and make it work

svetlemodry:
Oh, I see. Thanks!
And sorry, did not have to do much with .rst files so far.
Next time then .)

PR #933, opened 2021-08-29
Fix flatpak note to installing modules with pip3

### LINKED ISSUE #104: site https://qgis.org// | '--command=pip3' | "pip3: No such file or directory"
**Reference:** [page](https://qgis.org/en/site/forusers/alldownloads.html#flatpak); **Object**: "`flatpak run --command=pip3 org.qgis.qgis install urllib3 --user`"

```
$ flatpak --command=pip3 run org.qgis.qgis --user install urllib3
bwrap: execvp pip3: No such file or directory
```

aleixpol:
--devel needs to be passed as it comes from the SDK.

Fixes https://github.com/flathub/org.qgis.qgis/issues/104

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -495,7 +495,7 @@ Flathub files: https://github.com/flathub/org.qgis.qgis and report issues here: 
 
 Note: if you need to install additional Python modules, because they are needed by a plugin, you can install the module with (here installing the urllib3 module)::
 
- flatpak run --command=pip3 org.qgis.qgis install urllib3 --user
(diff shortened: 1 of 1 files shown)


aleixpol:
`https://github.com/flathub/org.qgis.qgis/blob/master/README.md` we'd always described it like this.

Yes, --devel is to pull the contents of the SDK, where pip3 is, so you need to install org.kde.Sdk.

If we were to consider that pip is an integral part of QGIS's UX, we could include it in the bundle. Not sure it's a great idea though.

rduivenvoorde:
Thanks! Sorry for leaving this open so long..

PR #937, opened 2021-10-06
Add 3.20 changelog to tx config file

DelazJ:
I think there is more to do (issue: 3.20 translations are not available in builds) but I forgot the steps ;( and the readme does not help me that much

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -907,6 +907,12 @@ source_file = i18n/en/LC_MESSAGES/site/forusers/visualchangelog318/index.po
 source_lang = en
 type = PO
 
+[qgis-website.site_forusers_visualchangelog320_index]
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
Hi @DelazJ the 'instructions' are here:

https://github.com/qgis/QGIS-Website/blob/master/Makefile#L183-L191

I did that (most important is the scripts/pre_... script) and updated the .tx/config
(and indeed only the 3.20 was to be added, normally I do that, as part of the creation of the visual changelog, apparently I forgot, or asked people to check it and then forgot)...

rduivenvoorde:
Thanks for noting and taking action by the way !!!!

DelazJ:
Thanks @rduivenvoorde 
I did also check the makefile but obviously too fast.
It could be nice that we harmonize processes/instructions at least between the docs and the website (by the end of next year? expecting the site gets updated in the meantime?)

Issue #952, opened 2021-11-22
Confusion about the LTR release


wonder-sk:
The situation about the current status of LTR is confusing for users:

The qgis.org website header suggests that 3.16.13 is the latest LTR:

![image](https://user-images.githubusercontent.com/193367/142873296-866aedd5-36ae-4511-a69f-86155e86e3b8.png)

Yet the blog on the front page says 3.16.13 was reverted to 3.16.11:

![image](https://user-images.githubusercontent.com/193367/142873157-4fc413f7-168b-4202-aea7-1e6aa7580d67.png)

But downloads offer 3.16.13.

In addition to that, the "LR" and "LTR" columns on the roadmap page seem to be the other way around:

![image](https://user-images.githubusercontent.com/193367/142873695-bd7e8135-9a85-40bf-8a1a-25e34f9938eb.png)

jef-n:
The (windows standalone) packages for 3.16.13 were updated and the packages for 3.16.14 are being build.

hannesaddec:
Hi, 
1) I'm still confused about LTR and RECENT version in the top links at qgis.org website. Should it not be 3.22 **LTR** and 3.16.13 for recent (so no extra characters)
2) which LTR is valid at the moment ?

jef-n:
3.22 is the current release and also a LTR release (ie. gets continued long term maintanance).  3.16 is also LTR and is currently in the LTR package repositories and gets further support until 3.22 is replaced by a newer release.   Then 3.22 will also replace 3.16 in the LTR package repositories.

johnwbryant:
I'm also confused by the headings on the [road map page](https://www.qgis.org/en/site/getinvolved/development/roadmap.html) as @wonder-sk mentioned. It feels like these might be reversed?
![image](https://user-images.githubusercontent.com/1100600/145336155-88fb2db0-d4b2-4a38-a137-7eb8f0cc227a.png)

johnwbryant:
Apologies for bringing it up one more time, I'm just trying to understand because I use the road map to explain it to QGIS learners in training sessions. 

The website indicates that in Feb 2022 the Latest will be 3.22.4 and the Long-Term Repo will be 3.24.0, which is the opposite of what I've been teaching.

![image](https://user-images.githubusercontent.com/1100600/145495309-6fff7ca5-aca8-492f-aa47-d6a28096089d.png)
[...]

PR #955, opened 2021-12-09
Display correct header for releases repo in road map

DelazJ:
Refs https://github.com/qgis/QGIS-Website/issues/952#issuecomment-989510727

@jef-n?

### DIFF
--- a/source/site/getinvolved/development/schedule.inc
+++ b/source/site/getinvolved/development/schedule.inc
@@ -5,7 +5,7 @@
 .. role:: rm-future
 
 .. csv-table:: Schedule
-   :header: "Event", "Latest", "Long-Term Repo", "Freeze", "Date", "Week #", "Weeks"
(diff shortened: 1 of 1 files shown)


jef-n:
Latest stands for the latest release an always goes to the first column.  The second column is about the LTR repos that get the LTR versions, once they are not the current release anymore - meanwhile the previous LTR is still continued in the LTR repos.  So the headings are correct.

jef-n:
ouch, if I hadn't messed up the column order in 36d8f4e73620a8

johnwbryant:
@jef-n are you sure it's right? Latest = 3.22.4 and Long-Term Repo = 3.24.0 in Feb 2022?

https://github.com/qgis/QGIS-Website/issues/952#issuecomment-990433416

DelazJ:
@johnwbryant you are right and this is corrected. the site just needs a rebuild

johnwbryant:
Phew - thought I was losing it! Thank you :)

PR #967, opened 2022-02-01
Fix documentation for installing pip3 modules

aleixpol:
We need to pass --devel otherwise pip3 is not found.


agiudiceandrea:
@aleixpol, duplicate of #933 just merged?

rduivenvoorde:
@agiudiceandrea ah yes, I think so. I'll close this one.

aleixpol:
Yes, it took so long to merge that I thought someone might have rewritten the text. I should have checked if it had been ignored, sorry.

Issue #879, opened 2021-03-21
Legacy release download address for MacOS


BerginJay:
## Description

Open the following URL, and click `All RELEASE` botton. 

Page URL: https://qgis.org/en/site/forusers/download.html

The download URL for OS X in `Previous releases of QGIS are still available here - including older releases for OS X here` is:

https://qgis.org/downloads/macOS/

It contain QIGS version from 2.14 to 3.8. However, https://qgis.org/downloads/macos/ contains more recent versions of QGIS.

So I suggest that we should merge these two or change link address in `Previous releases of QGIS are still available here - including older releases for OS X here`.

DelazJ:
See also #463 for Windows releases

BerginJay:
So is this a feature?

kyngchaos:
I think what happened here is that, after a long time of releasing the Mac package on my website, I made it available directly at QGIS.org in the macOS folder (that's why there are only a couple years of releases there).  Then I passed the "official" build to Petric as an all-in-one package and he created a new folder with all lower case, macos.

So, the dual folder setup just happened.  It may be a good idea to move the older packages into the current folder as a subfolder.  I don't think there any internal links that will break, but there may be external links to these old packages.  I do keep a full archive of my old packages on my site.

BerginJay:
Thanks! Since it has been fixed in #997, I now close this issue.

PR #997, opened 2022-05-13
Fix MacOS archive link

spatialthoughts:
Change the link from https://qgis.org/downloads/macOS/ to https://qgis.org/downloads/macos/

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -504,7 +504,7 @@ A nightly updated standalone installer from QGIS master can be downloaded from `
 
 Old releases
 ------------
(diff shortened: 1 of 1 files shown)


spatialthoughts:
This is a small but important fix. Many folks struggle to find the correct archive repo due to this typo. Gentle reminder to review and merge. Thanks!

andreasneumann:
Thank you @spatialthoughts

agiudiceandrea:
I think it would be better to link both https://qgis.org/downloads/macOS/ and https://qgis.org/downloads/macos/.

DelazJ:
Agreed (see #879) but in the meantime better advertise the macos url

PR #1003, opened 2022-06-20
Changelog 3.26: minor edits/additions related to point clouds

wonder-sk:
- add entry about classified renderer improvements for point clouds
- move an entry about following 2d renderer by 3d renderer for point clouds to the "Point Clouds" category where it better fits


rduivenvoorde:
Thanks Martin!

PR #1014, opened 2022-07-26
add quick start to readme

PeterPetrik:
I was struggling to get webpage with the styling up and running, so after a hint from @rduivenvoorde, this one works nicely on my PC


rduivenvoorde:
Thanks Peter!

PR #1017, opened 2022-08-02
Use code-block instead of bactick for code samples

DelazJ:
for a better rendering of https://qgis.org/en/site/forusers/visualchangelog324/index.html#feature-parameter-values-for-qgis-process-can-be-specified-as-a-json-object-via-stdin


rduivenvoorde:
Thanks!

Issue #960, opened 2022-01-10
Download page - Outdated links for OpenSuse. OpenSuse Leap 15.3 is missing and repository contains only qgis-ltr.


TellingBerto:
Hello,

download page for QGIS does not provide correct information w.r.t. openSuse Leap.

Repository for OpenSuse Leap 15.1 has been removed by openSuse.
Repository for OpenSuse Leap 15.3 is missing as link.
Repository for OpenSuse Leap 15.3 does provide only qgis-ltr. 
This is not mentioned on the download section for the QGIS-Website.
Feature or bug?

Thanks
T. Bertolli

jef-n:
We don't provide own openSUSE packages.  No sure who is maintaining those packages.   @kalxas @dassau

kalxas:
Leap 15.3 is still on Python 3.6 so the QGIS build never starts
https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/OSVNT77IXIHDHMKUZGMSEV7NYFXLHUKM/

![image](https://user-images.githubusercontent.com/383944/148758926-fa7410f2-71f9-4f96-acb2-95e441ce429a.png)

PR #1018, opened 2022-08-02
Fix Linux links in Download page

agiudiceandrea:
Should the openSuse info be removed? See: https://github.com/qgis/QGIS-Website/issues/960


rduivenvoorde:
Thanks!

PR #1031, opened 2022-08-19
publish to github pages

m-kuhn:


timlinux (approved):
LGTM

m-kuhn:
@timlinux building the website now should be working. Deploying to github pages fails, very likely because of the size of the whole site which is currently built for ~25 languages at ~560MB each (mostly changelog pictures) ==> ~15G.

Possibilities:
- Deduplicate pictures (needs understanding of the sphinx workflow which I lack)
- Deploy to another infrastructure than gh pages

Something else to consider is that we currently have a permanent redirect (HTTP 301) from qgis.org to qgis.org/en/site, I am not sure if/how that could be done on github pages.

rduivenvoorde:
> Deduplicate pictures (needs understanding of the sphinx workflow which I lack)

@timlinux @m-kuhn note that current reason the pictures are NOT deduplicated is that it now is possible that languages like japanese/arabic can create their own screendumps (so they can actually read them :-) )

But to be honest, nobody does that...

And we could still make that work (technically) by keeping current image-directories, but instead of copies of the images use (linux)symlinks IF the english image == <lang> image. That would be some script going over all image dirs and clean up.

Also I recall (@SrNetoChan @DelazJ did you find that out?) that (newer versions of) sphinx changed the handling of images, and instead of an image dir for every language there was now some option to add the lang to the image.
Something like: icon.png would have been replaced in the Dutch build if there is also a icon_nl.png image.
But I'm not sure of that.

I'm fine with building on github/actions, but isn't it better to keep serving themselves? Don't you think at a certain time github/micro$oft will stop doing that for free?
Would it be an option to build via an action but then rsync them to a webserver (backed by cloudfront, although off course probably also will stop doing stuff for free some day)?

.github/workflows/pages.yml
```diff
@@ -73,11 +43,8 @@ jobs:
       - name: Adding Known Hosts
         run: ssh-keyscan -H ${{ secrets.WEBSERVER_IP }} >> ~/.ssh/known_hosts
       
-      - name: hardlink duplicate files
-        run: jdupes -Lr output
-        
       - name: Deploy with rsync
-        run: rsync -avz output/html ${{ secrets.WEBSERVER_IP }}:${{ secrets.WEBSERVER_ROOT_FOLDER }}/${{ matrix.language }}
+        run: rsync -avz output/html ${{ secrets.WEBSERVER_IP }}:${{ secrets.WEBSERVER_ROOT_FOLDER }}/
```
m-kuhn: @jef-n it seems we need to specify the user here, it's currently connecting as runner@ which is not the user for which the ssh key is configured on the server. @timlinux should know more

jef-n: > @jef-n it seems we need to specify the user here, it's currently connecting as runner@ which is not the user for which the ssh key is configured on the server. @timlinux should know more

Thanks.  The receiving user is runner now too.  Just the build ran out of diskspace…  If 13bf209 doesn't help, I'll try sequential builds…

timlinux:
> I'm fine with building on github/actions, but isn't it better to keep serving themselves? Don't you think at a certain time github/micro$oft will stop doing that for free?
> Would it be an option to build via an action but then rsync them to a webserver (backed by cloudfront, although off course probably also will stop doing stuff for free some day)?

@rduivenvoorde thanks for the comments. Yes we will be pushing to a hetzner cloud server - mainly because the built assets exceed what we can host on gh pages even after using jdup. I updated the notes in the README (see PR #1039 ) with details of how I set up the server at the bottom of the README.

After we have it running there can we decommission the www2 server which I think is a more power hungry machine than we need?

.github/workflows/pages.yml
m-kuhn: Thanks for the runner.

The parallel approach before was working fine I think. It was a problem with gh pages where we needed to assemble one monolithic tar but rsyncing as now should be fine.

rduivenvoorde:
> After we have it running there can we decommission the www2 server which I think is a more power hungry machine than we need?

www2 is relative light cloud server with a bigger disk because it serve's (only, building is done on another) all (old doc (html/zipped/pdf) and website) pages:

```
# ls -l
total 60
lrwxrwxrwx  1 root root   33 Dec  6  2017 2.8 -> QGIS-Documentation-2.8/live/html/
drwxrwsr-x  3 1001 1008 4096 Jun  4  2020 doc-test
drwxrwsr-x  3 1001 1008 4096 Apr  8  2019 QGIS-Documentation
drwxrwsr-x 13 1000 1008 4096 May 14  2014 QGIS-Documentation-1.8
drwxrwsr-x 12 1000 1008 4096 Apr  5  2014 QGIS-Documentation-2.0
drwxrwsr-x  3 1001 1008 4096 Dec  1  2017 QGIS-Documentation-2.14
drwxrwsr-x  3 1001 1008 4096 Apr  8  2019 QGIS-Documentation-2.18
drwxrwsr-x  3 1000 1008 4096 May 20  2014 QGIS-Documentation-2.2
drwxrwsr-x  3 1000 1008 4096 Dec  2  2014 QGIS-Documentation-2.6
drwxrwsr-x  3 1000 1008 4096 May 22  2015 QGIS-Documentation-2.8
drwxrwsr-x  3 1001 1008 4096 Mar 15  2020 QGIS-Documentation-3.10
drwxrwsr-x  3 1001 1008 4096 Oct 30  2020 QGIS-Documentation-3.16
drwxrwsr-x  3 1001 1008 4096 Nov  2  2021 QGIS-Documentation-3.22
drwxr-xr-x  3 1001 1008 4096 Apr  8  2019 QGIS-Documentation-3.4
drwxrwsr-x  5 1001 1008 4096 Oct 14  2020 QGIS-tests
drwxrwsr-x  3 1001 1008 4096 Dec  4  2017 QGIS-Website

/var/www# du -hs qgisdata/
131G    qgisdata/
```

The one github is building on now, is ONLY the website...?

I'd say we should build to www2?

m-kuhn:
> The one github is building on now, is ONLY the website...?

So far, yes

.github/workflows/pages.yml
jef-n: Right, when I started I didn't know where the upload was actually pointed at and thought we were still going for gh pages and needed to reduce the volume - which doesn't work if we build each language in a separate job (at least not with jdupes).

PR #1036, opened 2022-08-20
Adds iOS in all_downloads pages

haubourg:
keeps consistency with main download page that had iOS added 1 month ago


haubourg:
Commits should be squashed probably. If this can't be done on merge in onee lcick, let me know and I'll make a cleaner PR

Issue #858, opened 2021-01-15
Website could guide people downloading QGIS more consistently


AlisterH:
## Description

Page URL: https://qgis.org/en/site/forusers/alldownloads.html
1. This page says "For new [windows] users we recommend the standalone installers."
But the [main installers page](https://www.qgis.org/en/site/forusers/download.html) has the OSGeo4W installers above the standalone installers.  So if users get there without reading the alldownloads page they are most likely to download the OSGeo4W installer as it is the first option they see.
If the recommendation on the alldownloads page is right, the main installers page should provide the standalone installers above the OSGeo4W installers.  Otherwise the recommendation should be amended.

2. The Windows section has a "Note the warning" link at the bottom, but it seems to have lost its context.  There is also a "(note the warning)" link in the main "QGIS installers" section and the Linux section, but these are in context (they are each at the end of a sentence about installing testing builds).  To give it some context, should the "Note the warning" text in the Windows section be expanded (e.g. "Before installing any of the nightly builds note the warning"), or perhaps made into another footnote of the table above it?

rduivenvoorde:
@AlisterH I agree on both of you points. Can you do a PR yourself? Or should I just do it.
About point 2, maybe make the 'note the warning' part of the 'versions table just above it' in the lines about development version?

jef-n:
I'd recommend the OSGeo4W installer for regular users.

rduivenvoorde:
@jef-n personally, me too. But that is not what the site tells us, AND (I think) the majority of Windows users are able to use...

jef-n:
It recommends the standalone installer to new users only.

AlisterH:
See pull request #886
Re point 1, I have just added hints rather than changing the order as I originally suggested.  Hopefully seeing "recommended for regular users" will prompt new users to keep reading and see the "recommended for new users".  Or do you guys think thIs is a cop-out?

DelazJ:
Anything left or can we close this report?

PR #1037, opened 2022-08-21
WIP: Download page simplification proposal

haubourg:
Hi all, 
following some discussions at the hackfest with @mbernasocchi @m-kuhn , I tried to patch the existing page to do a minor overhaul . 
The idea is to have a much simpler download page for windows with one main installer clearly visible. Blender download page is really nice for this :  

image

Here is a first comparison screenshot , I will keep on removing text and complexities, but we'll need a front end guru to polish this (hi @faebebin :wave:  ) if you think this is worth it. 

download_windows_option_1

A long running debate is wether we promote LTR or release first. 

My personal feeling is that Blender is right. New user just want one button pointing to the latest version. GIS admin doing deployment will probably take more time to read and the "Want more stability?" hooking formula sounds nice enough to me. 

any opinion @timlinux @andreasneumann ? 

related issues : https://github.com/qgis/QGIS-Website/issues/858

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -67,8 +67,8 @@
                             <li class="download-text">sha256</li>
                           </ul>
                         </a>
-                        <p>{{ _('Standalone installers are recommended for new users.') }} </p>
-                        <p><i>{{ _('Note that the MSI installers are much bigger than the previous installers. This is because they include significant larger packages (eg. PROJ 8).  The main reason for the switch to MSI were the size limits previously used NSIS has, which was blocking updates of dependencies.') }}</i><br/><br/></p>
+                        <p><i>{{ _('Standalone installers are recommended for new users.') }} </i></p>
+                      
                       </div>
                       <div>
                         <h6>{{ _('All in one OSGeo4W installer :') }}</h6>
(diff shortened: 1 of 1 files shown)


andreasneumann:
Thank you @haubourg -

Personally, I would get rid of the paragraph explaining why the MSI packager is so large. It has been a while like this now, and people probably don't remember the old times ...

On the other hand, people may not know what the "sha256" thingie means. Perhaps a small footnote (with an asterix) that explains in small font what it means?

faebebin:
@haubourg  Just stripped it down a bit more ...

image

haubourg:
> Personally, I would get rid of the paragraph explaining why the MSI packager is so large. It has been a while like this now, and people probably don't remember the old times ...

Yep, done 

image



> On the other hand, people may not know what the "sha256" thingie means. Perhaps a small footnote (with an asterix) that explains in small font what it means?

I would even be in favor of moving the cheksums in the all_downloads page and do not even bother with explaining them in the main page. 




> @haubourg Just stripped it down a bit more ...

yep yep, right direction here!

haubourg:
> Yep, done

 wrong screenshot sorry 

image

m-kuhn:
Can we just call it "Download QGIS 3.22"?
Ideally styled as a button?

jef-n (approved):

faebebin:
@haubourg I opened a final suggestion based on this PR here  #1046

haubourg:
> @haubourg I opened a final suggestion based on this PR here  #1046 

Cool! I'll have a look at it my way back from foss4g 😁

haubourg:
Closing this PR, superseded by #1046

PR #1044, opened 2022-08-24
Fix QGIS-windows-weekly paragraph

agiudiceandrea:
Fixes the QGIS-windows-weekly paragraph styling at https://qgis.org/en/site/forusers/alldownloads.html#windows which it seems to me it is currently wrong

image

### DIFF
--- a/source/site/forusers/alldownloads.rst
+++ b/source/site/forusers/alldownloads.rst
@@ -63,10 +63,10 @@ Steps are:
 
 .. _QGIS-windows-weekly:
 
-  For users that cannot use OSGeo4W for some reason or just prefer it there is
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
I was not aware of those weekly's.... That would have been the best answer when Bo asked for a way to run current master?

agiudiceandrea:
> I was not aware of those weekly's.... That would have been the best answer when Bo asked for a way to run current master?

@rduivenvoorde, they were reintroduced only yesterday with https://github.com/qgis/QGIS-Website/commit/5dce0197c431e21fee6a4db79448c46f28e47ad4.

PR #1046, opened 2022-08-25
Download page simplification

faebebin:
Based on #1037 

qgis-website-downloads

Also adapted the Max-OS section for convenience. Else the css would have to be doubled.

image

Moved the SHA256 links to all-downloads. Please check the .rst implementation , never worked with sphinx / .rst before.


source/site/forusers/alldownloads.rst
```diff
@@ -48,11 +48,11 @@ Steps are:
   +-------------------+------------------------------+------------------------+-------------------------------------------------------+
   | Release           | Version                      | Package                | Description                                           |
   +===================+==============================+========================+=======================================================+
-  | Latest Release    | |version|.x |codename|       | qgis                   | Release                                               |
+  | Latest Release    | |version|.x |codename|       | qgis                   | Release :windows_sha256:`[SHA256] <>`                 |
   |                   | |releasenote|                +------------------------+-------------------------------------------------------+
```
faebebin: please check whether this is true in all cases, as the jinja implementation was `release|e` instead of `version` etc ...

haubourg: @faebebin I have no clue about which variable to use. @jef-n or @rduivenvoorde do you know more about the above question?

haubourg:
Nice work Fabiano.  I streamlined again more, removing redundant informations. Should I send you a patch or submit a PR to your fork ? 
image

haubourg:
[patch.zip](https://github.com/qgis/QGIS-Website/files/9438202/patch.zip)

here is a good old patch :) (my train arrives at destination )

themes/qgis-theme/forusers_download.html
```diff
@@ -35,61 +35,38 @@ <h4>{{ _('Download for Windows') }} <i class="icon-chevron-down pull-right"></i>
...
+                        <p><i>{{ _('Standalone installers are recommended for new users.') }} </i></p>
+                      
+                      </div>
+                      <div class="card">
+                        <div>
+                          <a class="btn btn-large" href="https://download.osgeo.org/osgeo4w/v2/osgeo4w-setup.exe">
+                            <div class="primary-download-button">
+                              <div class="{{ "qgis{}-download".format( "2" if version.startswith("2.") else "" ) }}"></div>
+                              <div class="download-text">{{ _('OSGeo4W Network Installer') }}</div>
+                            </div>
```
m-kuhn: I'd suggest to stick to one single download button (and make the other one a link)

See suggestion below

faebebin: Thanks for improvements, wil have a close look soon

themes/qgis-theme/forusers_download.html
```diff
@@ -35,61 +35,38 @@ <h4>{{ _('Download for Windows') }} <i class="icon-chevron-down pull-right"></i>
...
-                              <li></li>
-                              <li></li>
-                              <li class="download-text">sha256</li>
-                          </ul>
-                        </a>
-			<p>{{ _('Note that the MSI installers are much bigger than the previous installers.  This is because they include significant larger packages (eg. PROJ 8).  The main reason for the switch to MSI were the size limits previously used NSIS has, which was blocking updates of dependencies.') }}</p>
+                        <p><i>{{ _('Standalone installers are recommended for new users.') }} </i></p>
+                      
+                      </div>
+                      <div class="card">
```
m-kuhn:
```suggestion
```

Should we keep it all on one card?

themes/qgis-theme/forusers_download.html
m-kuhn:
```suggestion
                          <a class="secondary-download-link" href="https://download.osgeo.org/osgeo4w/v2/osgeo4w-setup.exe">
                            {{ _('OSGeo4W Network Installer') }}
```

themes/qgis-theme/forusers_download.html
```diff
@@ -35,61 +35,38 @@ <h4>{{ _('Download for Windows') }} <i class="icon-chevron-down pull-right"></i>
...
-                              <li class="download-text">{{ _('QGIS Standalone Installer Version %s')|format( version ) }}</li>
-                          </ul>
-                        </a>
-                        <a href="https://qgis.org/downloads/QGIS-OSGeo4W-{{ release|e }}-{{ binary|e }}.sha256sum" class="reference external button">
-                          <ul class="inline download-row">
-                              <li></li>
-                              <li></li>
-                              <li class="download-text">sha256</li>
-                          </ul>
+                        <a class="btn btn-large" href="https://qgis.org/downloads/QGIS-OSGeo4W-{{ release|e }}-{{ binary|e }}.msi">
```
m-kuhn:
```suggestion
                        <a class="btn btn-large btn-success" href="https://qgis.org/downloads/QGIS-OSGeo4W-{{ release|e }}-{{ binary|e }}.msi">
```

+ change css for icon-large to a white logo (see https://qgis.org/en/_downloads/33f050ac04289004b33a4b296306d781/qgis-icon-white64.png)

m-kuhn:
image (16)

result of some playing here

m-kuhn:
We should also move the "Note" to sit at the bottom

haubourg:
mm, I like this greenish button. 

+1 for moving the note at the bottom, this was in my patch :)

jef-n:
Why move the SHA256 away from the standalone installers?  They are unrelated to the osgeo4w packages.

haubourg:
Well the feedback and discussion was to make the main download page as simple as possible. It is currently a bit crowded.  Sha checksums are useful in rare occasions. Having them in the advanced download page is probably enough. I know a lot of users that don't have a clue at all of what it is.

faebebin:
> Why move the SHA256 away from the standalone installers? They are unrelated to the osgeo4w packages.

was actually @andreasneumann suggestion. And I think majority agrees with it...

faebebin:
Thanks @haubourg  and @m-kuhn for improvements. I also applied the new download button to mac section (without any effort)  for following result:
image
image

haubourg:
Nice, did you apply my patch? I see some strings I removed. Alternatively, I'll open a PR to your branch, tell me what's better for you

faebebin:
I tried to apply but got an error on html generation. And then I just applied @m-kuhn suggestions.

faebebin:
Yes, open a PR

faebebin:
@m-kuhn , @haubourg , I applied the patch. Changes to windows download section from

image

to

image

haubourg:
Thanks @faebebin 

We need some spacing improvements I think. Can you merge "Standalone .. " and "NOTE: ..." into a single short line like 
`Standalone installers are recommander for new users (requires Windows >= 8. W7 not longer supports python 3.9 )`

And the link to the osgeo4w installer should be closer to its note.  
After this, I'll be quite happy :)


image

andreasneumann:
I think "not longer" should read "no longer" in the sentence "Python W7 not longer supports ..."

DelazJ:
And probably `Windows 7` instead of `Python W7`? Btw, is it not Python that dropped W7 support 😈?

DelazJ:
The note as in https://github.com/qgis/QGIS-Website/pull/1046#issuecomment-1229454434 looks nice to me...

jef-n:
> Btw, is it not Python that dropped W7 support 😈?

What do you mean?

DelazJ:
> What do you mean?

nitpicking: `W7 no longer supports python 3.9` vs `Python 3.9 (does not | no longer) support W7` are probably not the same thing.

jef-n:
> nitpicking: `W7 no longer supports python 3.9` vs `Python 3.9 (does not | no longer) support W7` are probably not the same thing.

Sorry - I missed the question mark and read "it is not Python that dropped W7 support".

faebebin:
I tried to implement all comments. I squashed and force pushed for easing conflict resolving.
(sidenote: Removed card border because makes no sense for single card)

@haubourg please reread the text content because againg ;) -

Would now look like this:

image

haubourg:
I like it. @m-kuhn @timlinux @mbernasocchi @jef-n @rduivenvoorde how does this main download page looks like to you?

jef-n:
> Sorry - I missed the question mark and read "it is not Python that dropped W7 support".

But speaking of which I found https://github.com/nalexandru/api-ms-win-core-path-HACK, which makes Python 3.9 work on W7 again (not sure to what extent, but seems to work) - it's also in OSGeo4W now (but nothing depends on it).

source/site/forusers/alldownloads.rst
jef-n: Please revert the changes to alldownloads.rst.    The checksums of the standalones don't belong there.

themes/qgis-theme/forusers_download.html
```diff
@@ -102,37 +71,34 @@ <h4>{{ _('Download for macOS') }} <i class="icon-chevron-down pull-right"></i></
...
+                            <div class="download-text">{{ _('QGIS macOS Installer Version %s')|format( version ) }}</div>
+                          </div>
+                        </a>
+                        <a class="secondary-download-link" href="https://qgis.org/downloads/macos/qgis-macos-ltr.dmg">
+                          {{ _('QGIS macOS Installer Version %s')|format( ltrversion ) }} LTR
+                        </a>
+                        <p><i>{{ _('Standalone installers are recommended for new users.') }} </i></p>
+                      </div>
+                      <div class="card">
+                        <h5>{{ _('Alternative build') }}</h5>
```
jef-n: The alternative builds are old.  They should be less emphasized than the LTR (or could IMHO just be removed as they are also mentioned on alldownloads)

jef-n (changes requested):

haubourg:
> > Sorry - I missed the question mark and read "it is not Python that dropped W7 support".
> 
> But speaking of which I found https://github.com/nalexandru/api-ms-win-core-path-HACK, which makes Python 3.9 work on W7 again (not sure to what extent, but seems to work) - it's also in OSGeo4W now (but nothing depends on it).

interesting. Embedding it in the installer would solve the issue ?  I'm a bit worried this could be spotted as something too hacky by antiviruses. As W7 is EOL do you think it is worth spending time on this ?

jef-n:
> interesting. Embedding it in the installer would solve the issue ? I'm a bit worried this could be spotted as something too hacky by antiviruses. As W7 is EOL do you think it is worth spending time on this ?

Who for some reason still wants/has to use W7, can install the package from OSGeo4W and hope for the best.  No sure what happens if you install it on >W7 - so I didn't plan on spending more time on this (or to put it into a standalone).

themes/qgis-theme/forusers_download.html
DelazJ: > could IMHO just be removed as they are also mentioned on alldownloads

Agreed

jef-n (approved):

faebebin:
@jef-n , @DelazJ , @haubourg , further simplified the download sections according above discussion:
* removed sha256 from alldownloads
* removed Mac OS alternative builds
* adapted and simplified Mac OS Section like Windows section

image

image

haubourg:
That's a big +1 for me too

m-kuhn:
If there are no objections, let's merge it this week

andreasneumann:
> If there are no objections, let's merge it this week

Personally, I would like the LTR to be advertised as the "default" and not the regular releases. They are more stable, offer a better experience for new users and it is the "de facto" standard in organizations and companies.

With the regular releases, esp. the .0 and .1 releases it is more of a "game" if they work or not.

m-kuhn:
Shall we do a voting?

React :heart: for regular release (3.26) and :rocket: for ltr (3.22) as main download

andreasneumann:
> Shall we do a voting?

Not sure if the github crowd is a representative group for the global QGIS user base ... but it is always good to "feel the pulse".

m-kuhn:
@andreasneumann a bigger redesign of the website is going on in parallel, so the decision here is only for the timeframe until that lands anyway (CC @timlinux )

jef-n:
@m-kuhn you can remove one of your votes ;)

haubourg:
As already shared with some of you, my point of view as a user and a former GIS administrator in charge of deploying QGIS for organisations, I vote to push the Release first. 
Why?
Because I think individual users are more seeking for the latest features and we need them to test it. So that LTR are more stable afterwards. And also because IT folks that choose to deploy a software know about LTS/LTR wording, when most users don't. They will be touched by the gentle nudge "Looking for more stable version?" . 

And in the end, if we are afraid that release are too much instable, we now have the "release candidate" wording in the splash screen and I hope no IT guy would deploy it with such a message. (I know, some will.. but hey... )

andreasneumann:
> Looking for more stable version?

But it doesn't read that way currently, it says "looking for long term support". It doesn't tell the users that LTRs are more stable.

anitagraser:
> I think individual users are more seeking for the latest features and we need them to test it. 

I wonder how useful it is to "trick" users into being testers. I think involuntary testers are unlikely to report bugs. At most, they complain about "the buggy os software" on social media.

haubourg:
> I wonder how useful it is to "trick" users into being testers. I think involuntary testers are unlikely to report bugs. At most, they complain about "the buggy os software" on social media.

Oh, sorry I didn't mean that we want to trick users into testers. Not at all.
Does this mean we are not confident enough in our releases?  I mean, ArcPro memes about freezes and crashes are legion. 
To me, the "release early , release often" philosophy is a strong point of open source software and diverting the biggest user base to the old versions is synonym to avoid benefiting of this advantage. 

Another way of saying this:  I tried for a long time now to convince organisations to use stabilized LTR's and use the small remaining time frame to have bug fixed through support contracts. But in reality, only a small fraction plays the game, which means the biggest user base is not helping us into stabilizing.  

When I download any other software, I expect the latest version first and I struggle to find examples where project pushes older versions first. Why would QGIS.org do that? 
Distributions on Linux play that game of selecting stable or bleeding age versions, not the software themselves.  


> But it doesn't read that way currently, it says "looking for long term support".

totally, we can rephrase this to "Looking for more stable version?" if it looks clearer . But it would looks a bit negative to me

m-kuhn:
> Does this mean we are not confident enough in our releases? I mean, ArcPro memes about freezes and crashes are legion. To me, the "release early , release often" philosophy is a strong point of open source software and diverting the biggest user base to the old versions is synonym to avoid benefiting of this advantage.

This is the key point to me. If we don't trust our releases we should take a step back and think what we can do to either improve stability or change the label. But this seems out of scope for the discussion here.

To me: we currently call them release so I am comfortable with offering them for downloading.

> > But it doesn't read that way currently, it says "looking for long term support".
> 
> totally, we can rephrase this to "Looking for more stable version?" if it looks clearer . But it would looks a bit negative to me

"Looking for the most stable version"?

andreasneumann:
> This is the key point to me. If we don't trust our releases we should take a step back and think what we can do to either improve stability or change the label. But this seems out of scope for the discussion here.

Right - personally I think we need to improve the semi-automated testing before we really recommend the regular releases over LTR releases. It happened quite a few times that the .0 or .1 had major issues (like a broken library in the packaging, or a major crash or regression).

@SrNetoChan and @gioman wanted to introduce the semi-automated testing, but I believe the process is not yet established. Once that is done, I think we are in a much better position to avoid major pitfalls for packaged releases.

themes/qgis-theme/forusers_download.html
```diff
@@ -35,62 +35,26 @@ <h4>{{ _('Download for Windows') }} <i class="icon-chevron-down pull-right"></i>
...
+                      <a class="btn btn-large btn-success" href="https://qgis.org/downloads/QGIS-OSGeo4W-{{ release|e }}-{{ binary|e }}.msi">
+                        <div class="primary-download-button">
+                          <div class="{{ "qgis{}-download".format( "2" if version.startswith("2.") else "" ) }}"></div>
+                          <div class="download-text">{{ _('QGIS Standalone Installer Version %s')|format( version ) }}</div>
+                        </div>
+                      </a>
+                      <a class="secondary-download-link" href="https://qgis.org/downloads/QGIS-OSGeo4W-{{ ltrrelease|e }}-{{ ltrbinary|e }}.msi">
+                        {{ _('Looking for Long Term Support ? Get QGIS Version %s')|format( ltrversion ) }} LTR
+                      </a>
+                      <p>{{ _('Standalone installers are recommended for new users. <em>Requires Windows 8 64 bit or higher because QGIS needs Python 3.9, which does no longer support Windows 7.') }} </em></p>
```
nyalldawson:
```suggestion
                      <p>{{ _('Standalone installers are recommended for new users. <em>Requires Windows 8 64 bit or higher because QGIS requires Python 3.9 or later, which is not available for Windows 7.') }} </em></p>
```

m-kuhn:
@andreasneumann are you ok if this is merged -- to make the download page look better in the short term -- and to follow up with the discussion about quality improvement somewhere else?

andreasneumann:
> @andreasneumann are you ok if this is merged -- to make the download page look better in the short term -- and to follow up with the discussion about quality improvement somewhere else?

@m-kuhn  Yes - I am fine with the change. Thanks!

agiudiceandrea:
@jef-n, it seems to me the main download page and the "all download" page now miss the previously available instructions on how to correctly install the full QGIS and QGIS LTR using the OSGeo4W Network installer:

> In the installer choose Express Install and select QGIS to install the latest release or QGIS LTR to install the long term release.
> The express installations have several optional packages including non-free software. To avoid those you have to use the Advanced Install and choose qgis and/or qgis-ltr in the desktop section.

agiudiceandrea:
Hi @faebebin @m-kuhn, it seems to me this PR has an unintended and unwanted side effect: please see https://github.com/qgis/QGIS-Website/issues/1077.

faebebin:
@agiudiceandrea  thanks for info. Will have a look asap

PR #1051, opened 2022-09-11
Fix ltr download link

m-kuhn:


agiudiceandrea:
Hi @m-kuhn and @jef-n, now that "'QGIS Standalone Installer Version" was replaced by just "Download QGIS", I think the following phrase "The standalone installers are recommended for new users" makes no sense for new users as there is no clue on what are the standalone installers.

PR #1052, opened 2022-09-11
Put legacy note in small print

m-kuhn:


jef-n:
I'd leave it as is - the note was moved below as it applies to both options (standalone / OSGeo4W installer) - it's small it might be overlooked esp. for the more prominent placed standalones at the top.

m-kuhn:
Support for Windows 7 and Windows 8 has reached EOL, does this apply to a platform that is still actively supported?

jef-n:
> Support for Windows 7 and Windows 8 has reached EOL, does this apply to a platform that is still actively supported?

Windows 7 is apparently still used. It didn't take long from taking down the earlier note to the first question why it didn't work on Windows 7 or whether it's 64 or 32bit.

m-kuhn:
I'd still assume that almost everyone visiting this page is not interested in Windows 7 and that this note should be big enough to be readable by the few people looking for it but small enough not to be auto-discovered by the majority.

PR #1061, opened 2022-09-14
Small typo's

DiGro:
Line 89 and 90: "There are also metapackages with postfix "-full-free" and "-full" with install the corresponding QGIS version and a additional optional packages that are"

corrections :  "with install "_> "which install" and "a additional" -> "additional"

should probably be": " There are also metapackages with postfix "-full-free" and "-full" which install the corresponding QGIS version and additional optional packages that are"


jef-n:
The passage was replaced in ab366b843c92f570ba2c59c21ebfecc05ebbabda

PR #1069, opened 2022-10-06
Fix matrix room link

strk:
Reported by mkoniecz via IRC (could not join Matrix otherwise)


strk:
Too fast, the link is not rendered correctly, when seen rendered by github at least: 
https://github.com/qgis/QGIS-Website/blob/9e4860691ea71ec2d1d333b40fed4590ee59afbd/source/site/forusers/support.rst

PR #1076, opened 2022-10-26
Add expression editor close safeguard to changelog

kannes:


kannes:
I had missed to add this to the changelog earlier but would love to see it included :)

The image filename is the sha1sum as the other images seem to use as well. The image is compressed with optipng. I hope I managed to write correct ReST syntax...

rduivenvoorde:
Thanks, will check

PR #1079, opened 2022-11-06
Fix (followup thankyou message) new tab for osgeo installer

### LINKED ISSUE #1077: Missing Thank You / Donation overlay text in the Download page
## Description

The Download page used to display an overlay with thanks and a Donation button after the user clicked on a download link.

image

Now such overlay is no longer displayed.

It seems to me this is an unwanted side effect of the "Download page simplification" PR https://github.com/qgis/QGIS-Website/pull/1046.

Looking at the qgis-donations bot activity https://github.com/qgis-donations, such side effect may be responsible for the decrease in donations reported by the qgis-donations bot after the 9th of September (when the "Download page simplification" was merged).

image

See also the thread at https://lists.osgeo.org/pipermail/qgis-community-team/2022-November/006083.html.

Page URL: https://qgis.org/en/site/forusers/download.html

faebebin:
Fixes #1077 (followup)

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -46,7 +46,7 @@
                         {{ _('Looking for the most stable version? Get QGIS %s %s')|format( ltrversion, ltrnote ) }}
                       </a>
                       <br/>
-                      <a class="reference secondary-download-link" href="{{ pathto('site/forusers/alldownloads') }}#osgeo4w-installer">
+                      <a class="reference external secondary-download-link" href="{{ pathto('site/forusers/alldownloads') }}#osgeo4w-installer">
                         <div class="primary-download-button">
                           <div class="osgeo-download"></div>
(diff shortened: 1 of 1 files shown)


faebebin:
@agiudiceandrea opens the link now in new tab, like before. Thankyou for correction.

image

It also adds this little `external arrow icon` . I did not put any effort into making improving positioning or so. Let me know if we want this or not...

BTW for the linux download links there is and was no thankyou message. Should we consider also adding it there, as we are already working on it?

agiudiceandrea:
Great! Thanks again.

> It also adds this little `external arrow icon` . I did not put any effort into making improving positioning or so. Let me know if we want this or not...

I think it is not a priority, but would be also good to fix it if possible without too much effort.

> BTW for the linux download links there is and was no thankyou message. Should we consider also adding it there, as we are already working on it?

I think it would be a good improvement.

faebebin:
Ok, will implement this and reopen the PR

PR #1084, opened 2022-11-11
Add Windows 7 latest easily installable version

DelazJ:
Can someone confirm that the max a Windows 7 user can install is 3.20? Thanks


nicogodet:
AFAIK, it's 3.18.3 with .exe installer (https://download.qgis.org/downloads/QGIS-OSGeo4W-3.18.3-1-Setup-x86_64.exe)

jef-n:
OSGeo4W now always installs the hack with python (see https://github.com/jef-n/OSGeo4W/commit/8f369ebea9f71443a7b0038e553ca195b9eb063c; only enabled on Windows <8).   So QGIS from OSGeo4W should also work on Windows 7 again (and coming standalone installers will also pick that up)

DelazJ:
So this whole note should actually be removed soon...

PR #1085, opened 2022-11-12
Fix link to "QGIS Application Network Connections" doc

agiudiceandrea:
"Testing" docs are only in English.


DelazJ:
Sorry if I'm late here. Maybe should we instead use latest, and not testing?

agiudiceandrea:
> Maybe should we instead use latest, and not testing?

It is an option, but if the testing is the most up-to-date page, then I think it is better to link to the testing page.

DelazJ:
This is a topic that is not related to a particular QGIS version or the docs CI/CD, so unless we forget, I do not see a particular reason the occasional changes would not be backported.

Issue #1089, opened 2022-12-05
The "Push strings to transifex" action is continuosly failing since Nov, 17th


agiudiceandrea:
## Description

See https://github.com/qgis/QGIS-Website/actions/workflows/tx_push.yml

DelazJ:
Looks like we need a tx token, hence someone with admin rights on this repo... @m-kuhn can you give it a look please?

m-kuhn:
The API version 2 on transifex was recently phased out, right now only API version 3 is available if I'm not wrong. https://pypi.org/project/transifex-client/ is not compatible with the new version and a different client need to be used instead.

agiudiceandrea:
Any news about this issue which is probably the root cause of the issue https://github.com/qgis/QGIS-Website/issues/1092?

m-kuhn:
I've tried to change the toke with no success.
@why-not-try-calmer could you have a look at this?

DelazJ:
@m-kuhn Have you tried to setup [a github<->transifex integration ](https://help.transifex.com/en/articles/6265125-github-via-transifex-ui). We have this implemented for the docs ([here](https://github.com/qgis/QGIS-Documentation#translating) and [here](https://www.transifex.com/qgis/qgis-documentation/settings/integrations/) and push and pulls seem to work. But TBH I no longer understand the logic of the website translation system so maybe my suggestion is useless.

m-kuhn:
No, I have no experience with that.

why-not-try-calmer:
Thanks for meging @DelazJ, to avoid the multiplication of PRs for fixing Transifex workflows I'll open one last PR to host all future commits related to the fix if that's ok for you. I'll mark it as not-to-be-merged-until-everything-works.

DelazJ:
@why-not-try-calmer I'm fine with whatever suits you. Just do, and tell us when you think it is ready.

why-not-try-calmer:
Tracking efforts & improvements in #1111. This issue could be closed.

DelazJ:
Let's then follow #1111

PR #1097, opened 2023-01-05
Up to date transifex client

why-not-try-calmer:
This PR aims at fixing the problem raised in #1089. It replaces the deprecated Python client with the up-to-date binary executable. It also adds a _workflow_dispatch_ trigger to test the action. No other change should be required (i.e. the current token should be fine).

__To do (after this has received workflow approval)__
- [ ] test push
- [ ] test shell script adding resources

### DIFF
--- a/.github/workflows/tx_push.yml
+++ b/.github/workflows/tx_push.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 
 # Allow one concurrent deployment
--- a/.github/workflows/tx_push.yml
+++ b/.github/workflows/tx_push.yml
@@ -23,6 +23,11 @@ jobs:
         run: |
           pip install -r REQUIREMENTS.txt
 
+      - name: Install Transifex client
+        run: |
+          curl -OL https://github.com/transifex/cli/releases/download/v1.3.1/tx-linux-amd64.tar.gz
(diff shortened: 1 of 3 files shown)


DelazJ:
> To do (after this has received workflow approval)

@why-not-try-calmer I approved to allow further steps, without fully understanding the scope of your changes. I assume @m-kuhn has trust in you if he calls you.

why-not-try-calmer:
> > To do (after this has received workflow approval)
> 
> @why-not-try-calmer I approved to allow further steps, without fully understanding the scope of your changes. I assume @m-kuhn has trust in you if he calls you.

Hello @DelazJ , thank you for your trust! The scope of the submitted changes is to use the up-to-date Transifex client as a drop-in replacement for the deprecated client, adjusting CI where necessary. My goal is to make the workflows / actions trying to push to Transifex work for good (they have been failing for several months unfotunately).

Since there's a dispatch_workflow setting, may I suggest to test the changes by manually triggering the action? (if I am not mistaken, I cannot do that myself). Then if something does work I can use the output of the failed action to iterate over the changes and make further adjustments until it works!

Let me know, and in the mean time, have a pleasant day!

m-kuhn:
@why-not-try-calmer thank you
I think we need to first merge before we can use the dispatch action, or am I missing something?

This looks good to me for a test ride, worst case it just won't work

DelazJ:
+1 to merge and see

PR #1098, opened 2023-01-07
Update ukrainian_spt.rst

oleksaboiko:
Changed the scale of the pictures


DelazJ:
@oleksaboiko sorry for the late feedback. I couldn't get preview of images in the generated artifacts but I guess it is better than current situation.

PR #1102, opened 2023-01-23
Fixing Transifex translations

why-not-try-calmer:
Getting `create_translation_resources.sh` to work.


why-not-try-calmer:
Can a maintainer confirm that the organization (`o:`) and project (`p:`) match what you can see under your Transifex.com dashboard?

If this is the case, this can be considered merge-able from my viewpoint.

rduivenvoorde:
I'd say: yes?

Screenshot from 2023-01-30 10-16-19

why-not-try-calmer:
This should work now.

PR #1107, opened 2023-02-02
Addressing permission error on create_transifex_resources

why-not-try-calmer:


why-not-try-calmer:
@DelazJ Can we see if this worked? We recently 'fixed' 3 other repos doing a similar migration and I'd like to finish this while it's fresh on my mind.

m-kuhn (approved):

PR #1144, opened 2023-03-31
Update features.rst

claudiadragoste:
Add the link https://github.com/qgis/qwc2 for  "QGIS Web Client".

### DIFF
--- a/source/site/about/features.rst
+++ b/source/site/about/features.rst
@@ -40,7 +40,6 @@ powerful symbology, labeling and blending features to impress with your
 maps.
 
 `QGIS Web Client 2 <https://github.com/qgis/qwc2>`_
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. figure:: /static/site/about/images/qgiswebclientscreenshot.png
    :alt: QGIS Web Client
 
--- a/source/site/about/features.rst
+++ b/source/site/about/features.rst
@@ -40,6 +40,7 @@ powerful symbology, labeling and blending features to impress with your
 maps.
 
 `QGIS Web Client 2 <https://github.com/qgis/qwc2>`_
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


source/site/about/features.rst
```diff
@@ -32,7 +32,7 @@ QGIS server is considered as a reference implementation for WMS 1.3.
 
    A screenshot of QGIS Server configuration in QGIS Desktop
 
-QGIS Web Client
+`QGIS Web Client <https://github.com/qgis/qwc2>`_
 ---------------
```
DelazJ: This line should be as long as (or longer than) the title line above

claudiadragoste: I modified again the file. Is it OK like this?

source/site/about/features.rst
```diff
@@ -39,10 +39,12 @@ Publish your QGIS projects on the web with ease. Benefit from the
 powerful symbology, labeling and blending features to impress with your
 maps.
 
+`QGIS Web Client 2 <https://github.com/qgis/qwc2>`_
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
DelazJ:
```suggestion

```
The underline is then unnecessary here. And yes, I personally prefer adding the URL in the section (as you do now).

claudiadragoste: I agree, thank you

claudiadragoste: I preview without this line and it's not the same... (I tried to follow the exemple of "QField", "Mergin" or "IntraMap")

DelazJ: QField and others are subsections within the "QGIS on mobiles and tablets" section. Here, you already have a QWC section above  with some description and you want to add a single and eponym subsection. Doesn't look nice to me.

claudiadragoste: I understand. I did it in this way because I was thinking that "QGIS Web Client" could be the section; and "QGIS Web Client 2" - a subsection;  in case there are others qgis web clients. If you consider that is better to remove the subsection, I will do the changes

DelazJ: Just a link makes more sense to me currently. The current valid QWC is QWC2 so the same thing. and I don't think there will be more than one version at a time.

claudiadragoste: I made the changes.

PR #1151, opened 2023-04-18
Updated links for Datapolitan classes to new cont

richarddunks:
The classes were recently updated to fix broken links and missing content. The new links point to the new site for the materials on Github. Also updating the description to clearly state availability for self-study and re-use. 

Note: the short URL for the Intermediate class won't link to the right content until 1 May, but once that's fixed, it will point to this site: https://training.datapolitan.com/intermediate-gis/#1. Using the bitlinks will keep the links fresh so this shouldn't happen again.


rduivenvoorde:
@richarddunks why would a organisations want to change httpS links with secure certificates to anonymous http bitly links which can litterally link to everything (including the download of a trojan)?

Personally I would NOT link to bitly links, but othere might have less trouble with this.

agiudiceandrea:
@rduivenvoorde, it looks like this is superseded by https://github.com/qgis/QGIS-Website/pull/1238.

PR #1191, opened 2023-10-13
Chats are not just "IRC" chats nowadays

strk:
I went looking beacuse it surprises me no many active QGIS developers are in #qgis:osgeo.org ...


strk:
Thanks @rduivenvoorde

PR #1194, opened 2023-10-28
Prepare 3.34 docs release

DelazJ:
~~Probably not to merge yet !!!~~


agiudiceandrea:
It would be great if, just before the next time a new docs release, the text could be slightly modified in order to avoid to translate it each time. In particular the phrase: `We are still updating (not translating yet) the documentation for releases newer than QGIS 3.34. We call this version 'QGIS testing' and the documentation can be found here:`

DelazJ:
You mean , replace 3.34 by a variable ? which one? And I think we should do it now (I personally can't promise I'll remember this comment next autumn)

agiudiceandrea:
> And I think we should do it now

See: https://github.com/qgis/QGIS-Website/pull/1214

PR #1200, opened 2023-11-07
Update index.rst

EmHain8:
replaced Nyalls github with NR website and added in funder target for Cesium 3D tiles


rduivenvoorde:
Hi @EmHain8,

The QGIS changelog (or actually https://changelog.qgis.org, from which this rst is created) normally links to the persons github personal account, never to commercial entities.
Personally I prefer to keep it this way. Is that OK for you?
If not, can you please discuss this on the PSC mailing list?
I'm off course fine with changing the Cesium grant link.

nyalldawson:
Hey @rduivenvoorde 

> The QGIS changelog (or actually https://changelog.qgis.org/, from which this rst is created) normally links to the persons github personal account, never to commercial entities. 

This hasn't always been the case -- see eg https://qgis.org/en/site/forusers/visualchangelog326/index.html, which has links to multiple organisation sites.

> Personally I prefer to keep it this way. Is that OK for you?

Why is this though?  Why shouldn't the organisation employing the developer get credit (and potentially more work) for the contributions?

rduivenvoorde:
@nyalldawson I really thought we always did github accounts, I'm an idealist and still think these are individuals making the code changes, not the corporations. While they serve credits for paying the individuals, I personally prefer the qgis website NOT to be a commercial billboard for GIS corps. But I'm pretty alone in this :-)  apparently.
Feel free to merge.

agiudiceandrea:
Maybe this should be closed in favor of https://github.com/qgis/QGIS-Website/pull/1201.

PR #1209, opened 2023-11-28
Another attempt to have macOS users enable QGIS

### LINKED ISSUE #1195: Clarify instructions for QGIS download for OSX / Mac to bypass notarization issue
## Description

When installing QGIS on a Mac / OSX / MacOS you have to bypass the notarisation issue by using Control-Click and choose open on the menu. This is what we have on the page currently:

https://qgis.org/en/site/forusers/download.html

V0fJhpEDFByVjtW1

When working with novice users it often confuses them. I understand that QGIS is looking to notarise the installer to completely bypass the issue and I am not sure what the current status of this is. 
https://github.com/qgis/QGIS-Enhancement-Proposals/issues/218 
https://github.com/qgis/QGIS/issues/32167

In the mean time, I suggest updating the text on the website. I am working with a client who needed to install QGIS on a Mac, and got confused. I explained what need to be done, and asked them what changes they suggested to clarify this. 

He suggested this:

image002

"Were it me, I’d probably format it slightly differently (I’m drawn to the big green button and not to the stuff below it, personally). Quick mock-up below of what would probably made me read it, with some slight change on the focus of the wording. 

Hope this is useful; apologies if this is a bit OTT but I did a big UX project on exactly this sort of thing for a holiday rental business previously, so it’s sort of ingrained in me!"

Can we make these changes? What do you think?

Best wishes,
Nick.

DelazJ:
Hopefully fixes #1195
I made a kind of mix between @nickbearman feedback and Garth's one in the mailing-list pointed by @agiudiceandrea i.e.
move the opening information above the button (hopefully people read that before they click) and focus only on what they have to do (notarization issues is QGIS burden...)

### DIFF
--- a/themes/qgis-theme/forusers_download.html
+++ b/themes/qgis-theme/forusers_download.html
@@ -71,6 +71,7 @@
                   </div>
                   <div id="mac" class="accordion-body collapse in">
                     <div class="accordion-inner centered">
+                      <p><b><em>{{ _('After installing QGIS, the first launch attempt may fail due to security protections. To enable QGIS, control-click on its icon in your Applications folder and select Open in the context menu. A confirmation dialog will display where you you need to click the Open button again. This only has to be done once.') }}</em></b></p>
                       <a class="reference btn btn-large btn-success" href="https://qgis.org/downloads/macos/qgis-macos-pr.dmg">
                         <div class="primary-download-button">
                           <div class="qgis-download"></div>
@@ -81,7 +82,6 @@
                       <a class="reference secondary-download-link" href="https://qgis.org/downloads/macos/qgis-macos-ltr.dmg">
                         {{ _('Looking for the most stable version? Get QGIS %s %s')|format( ltrversion, ltrnote ) }}
                       </a>
-                      <p><b><em>{{ _('macOS High Sierra (10.13) or newer is required. QGIS is not yet notarized as required by macOS Catalina (10.15) security rules. On first launch, please control-click on the icon and choose Open from the context menu, after which a confirmation dialog is shown and you need to click the Open button.') }}</em></b></p>
                     </div>
                     <div class="accordion-inner">
                         <p>{{ _('Alternatively, for native support of both Intel x86 and Apple ARM architectures, together with the latest versions of key components (e.g. GDAL, PDAL, GRASS GIS) see:') }}</p>


themes/qgis-theme/forusers_download.html
nickbearman, line 1: Thanks @DelazJ . I'm not sure you have to go through Applications for this - it can be done from the icon on the bar at the bottom (from memory). 

How about:

After installing QGIS, the first launch attempt may fail due to security protections. To enable QGIS, control-click on its icon and select Open in the context menu. A confirmation dialog will display where you you need to click the Open button again. This only has to be done once.

DelazJ: @nickbearman Indeed, people don't need to go through Applications folder, but I don't think that macOS places icon of installed applications in the Dock by default. So the imho safest step is to ask them to use the Applications road. Also, Garth was mentioning that it would avoid people trying to ctrl-click the icon in the dmg file they are installing.

nickbearman: @DelazJ Good point, thanks for clarifying. Yes this looks good, thanks very much for making the change

agiudiceandrea:
@DelazJ, I think another issue now is that the message will result not translated in most languages.

DelazJ:
Yes but that is the "issue" whenever we adjust strings. Let's hope translators will take it quickly. Or if you feel like mentioning this in the qgis-tr list...
Moreover, if an issue now, this will probably be the permanent issue with the next "not-to-be-translated" website (assuming I understood correctly the orientation)

pathmapper:
Thanks for updating the instructions, just translated them to German.

> this will probably be the permanent issue with the next "not-to-be-translated" website (assuming I understood correctly the orientation)

Hmm, came along this ticket because a QGIS user contacted our user group regarding installation problems on Mac.

If the next website will not be translated, I'm afraid there will be many requests like these...

agiudiceandrea:
Another related issue report: https://github.com/qgis/QGIS/issues/55663.

pathmapper:
> the next "not-to-be-translated" website

According to https://docs.google.com/document/d/1eRPFRLGFcOtJzpxJz8Jk2UTPXJIrCRz2GCiJ0LfiBYY

> the new website will be in english only with an in-tandem investigation into using browser built in translation tools if possible

so there's a chance that there will be some translation also with the new website.

agiudiceandrea:
https://github.com/qgis/QGIS/issues/56149

PR #1220, opened 2024-01-20
Override Sphinx >=7.2.0 visited link color in carousel caption and button text

### LINKED ISSUE #1219: With the new Sphinx version the color of already visited links is different
## Description

Current home page rendering:
image

Previous home page rendering:
image

The change has been introduced in Sphinx 7.2.0 (https://github.com/sphinx-doc/sphinx/pull/11390)

I think it would be better to restore the previous behavior at least for the buttons text and for the links in the carousel.

agiudiceandrea:
I'm not completely sure this is the correct way to fix the issue, anyway it seems to me this fixes https://github.com/qgis/QGIS-Website/issues/1219.

This PR only overrides the Sphinx >=7.2.0 visited link color, restoring the previous behavior, only for carousel captions and for buttons text.


agiudiceandrea:
No longer applicable.

PR #1221, opened 2024-01-25
WIP: Add icons in roadmap table

strk:
I'm hoping to make it more readable (removing the needs to scroll up to find out which column represents what)

I'd like to also add icons for "Latest" (some bleeding edge thing) and LTR (some icon to represent stability) - ideas welcome :)

Should I also include changes to the .ics and .inc files in the same PR ?

\cc @jef-n


strk:
Obsoleted by #1222 which I also prefer

PR #1222, opened 2024-01-25
add icon to dev version (ref #1221)

jef-n:


strk:
I like this better and other can added I guess to the other two columns

Issue #1160, opened 2023-05-31
Information about the Project Organisation (Board, Community Voting Members) is not up to date


agiudiceandrea:
## Description

It seems to me the info in the [Board](https://www.qgis.org/en/site/getinvolved/governance/governance.html#board) and [Community Voting Members](https://www.qgis.org/en/site/getinvolved/governance/governance.html#community-voting-members) sections (perhaps other ones?) is not up to date.
It seems that also the name of the OSGEO QGIS voting representative is not mentioned.

agiudiceandrea:
Hi @mbernasocchi, have you seen this issue report?

mbernasocchi:
Hi @agiudiceandrea no i hadn't seen it. Thanks for pinging me. Will have a look tomorrow.

agiudiceandrea:
@mbernasocchi, any news? Please also have a look at the PSC meetings minutes archive page https://github.com/qgis/QGIS/wiki/PSC-Meetings-archive: the August, September, October and November meetings minutes are missing.

DelazJ:
> Please also have a look at the PSC meetings minutes archive page https://github.com/qgis/QGIS/wiki/PSC-Meetings-archive: the August, September, October and November meetings minutes are missing.

You beat me at reporting 😄

agiudiceandrea:
> @mbernasocchi, any news? Please also have a look at the PSC meetings minutes archive page https://github.com/qgis/QGIS/wiki/PSC-Meetings-archive: the August, September, October and November meetings minutes are missing.

The latest monthly meetings minutes are now no longer missing after https://github.com/qgis/QGIS/wiki/PSC-Meetings-archive/b04e4ecbae41f9ca84b44bce7a4e66ee6de3afdb

agiudiceandrea:
It looks like also the [Board](https://www.qgis.org/en/site/getinvolved/governance/governance.html#board) section has been updated recently with commit https://github.com/qgis/QGIS-Website/commit/68d74ab27f3c51a12c4119ee7031fc52b99e95c7.

It seems to me the [Community Voting Members](https://www.qgis.org/en/site/getinvolved/governance/governance.html#community-voting-members) section is still not up to date.

PR #1235, opened 2024-04-14
Update voting members

anitagraser:
I've added the community voting members that were added in 2021+22 as well as the OSGeo voting member

### DIFF
--- a/source/site/getinvolved/governance/governance.rst
+++ b/source/site/getinvolved/governance/governance.rst
@@ -27,7 +27,7 @@ Project Steering Committee
 --------------------------
 
 * Marco Bernasocchi (Chair) [Chair since May 2020; Vice chair April 2018 - April 2020]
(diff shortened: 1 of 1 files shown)


anitagraser:
Adresses https://github.com/qgis/QGIS-Website/issues/1160

agiudiceandrea:
Shouldn't the voting members elected to the PSC be removed from the voting members list?

https://qgis.org/en/site/getinvolved/governance/charter/index.html#voting-member-terms

> 5.6 Voting member terms
> Voting memberships will be valid until:
> 
> - The voting member resigns
> 
> - There is a motion raised from within the voting membership, and passed vote by the voting membership to remove the member
> 
> - <ins>A voting member gets elected to the PSC in which case voting membership of the new PSC member will be relinquished and a new voting member will be elected in the new PSC member’s place. Outgoing PSC members that were previously voting members will need to be re-elected into the voting membership should they wish to return to their former role</ins>.

PR #1236, opened 2024-04-15
Variable |yeartag| doesn't seem to render on page

DiGro:
On the page getinvolved/faq/index the variable |yeartag| doesn't seem to render.

Line 85 / 91 and a few more


agiudiceandrea:
@DiGro, probably I miss something... but it looks like there are no changes made with this PR.

DiGro:
@agiudiceandrea :  You are absolutely right :-) No change at all. I have absolutely no idea how to fix this.
I should have probably made an issue of this, so I will close this PR and make an issue instead

PR #1238, opened 2024-04-17
Update links to Datapolitan training materials

EmHain8:
links were pointing to an older version that was referencing QGIS 2.X and other outdated information

### DIFF
--- a/source/site/forusers/trainingmaterial/index.rst
+++ b/source/site/forusers/trainingmaterial/index.rst
@@ -32,8 +32,8 @@ http://mltconsecol.github.io/QGIS-Tutorial/
 
 In-person training material, originally developed for the New York City Department of Transportation by `Datapolitan <http://www.datapolitan.com>`_:
 
-* `Introduction to GIS Fundamentals <http://training.datapolitan.com/qgis-training/Introduction_to_GIS_Fundamentals>`_
-* `Intermediate GIS with QGIS and PostGIS <http://training.datapolitan.com/qgis-training/Intermediate_GIS/>`_
+* `Introduction to GIS Fundamentals <https://training.datapolitan.com/intro-gis-fundamentals/#1>`_
+* `Intermediate GIS with QGIS and PostGIS <https://training.datapolitan.com/intermediate-gis/#1>`_
 * `Source code for training materials <https://github.com/Datapolitan-Training/qgis-training>`_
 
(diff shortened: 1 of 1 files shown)


source/site/forusers/trainingmaterial/index.rst
```diff
@@ -32,8 +32,8 @@ http://mltconsecol.github.io/QGIS-Tutorial/
 
 In-person training material, originally developed for the New York City Department of Transportation by `Datapolitan <http://www.datapolitan.com>`_:
 
-* `Introduction to GIS Fundamentals <http://training.datapolitan.com/qgis-training/Introduction_to_GIS_Fundamentals>`_
-* `Intermediate GIS with QGIS and PostGIS <http://training.datapolitan.com/qgis-training/Intermediate_GIS/>`_
+* `Introduction to GIS Fundamentals <https://training.datapolitan.com/intro-gis-fundamentals/#1>`_
+* `Intermediate GIS with QGIS and PostGIS <https://training.datapolitan.com/intermediate-gis/#1>`_
```
DelazJ, line 36: @EmHain8 Afaict, this is full of QGIS 2 screenshots. No single QGIS 3 logo/dialog. Is this really the updated version?

EmHain8:
Ahh is it - sorry - we were just looking at it and it seemed better than what was linked previously. It should be an action to update it fully possibly. The value to updating to this version in the meantime takes out some of the problems with the previous version in some of the themes discussed that weren't suitable.

EmHain8:
I would say remove the link, but it is incredibly useful presentation.

DelazJ:
> we were just looking at it and it seemed better than what was linked previously. [...]
>The value to updating to this version in the meantime takes out some of the problems with the previous version in some of the themes discussed that weren't suitable.

Can you clarify? Any examples?
I walked through the first 50% pages of both versions of "intermediate lesson" and the only differences I saw in the new version is the font ( + colored background and title) and less clicks. Otherwise the text IS the same (except authorship on first page 👀 ).
I must admit however that the colors make the new version more appealing so if we update, it would be for that, not because it provides better/updated contents per say. let's just be clear on our intentions.

agiudiceandrea:
@EmHain8, does this PR supersedes https://github.com/qgis/QGIS-Website/pull/1151? If yes, should https://github.com/qgis/QGIS-Website/pull/1151 be closed?

DelazJ:
@agiudiceandrea Based on the description of that PR, yes it does!

PR #1241, opened 2024-04-27
Add Ecuador

Ta3344:


DelazJ:
@Ta3344 Ecuador is already listed. I guess this is a wrong PR...

PR #1196, opened 2023-11-04
Additions for Fedora, Opensuse and general Linux

boredsquirrel:
I started by adding up-to-date version infos for Fedora. Then I added the instructions for Toolbox and Distrobox, toolbox being standard on Fedora Atomic, Distrobox on Opensuse microOS but also working on Fedora Atomic. 

These tools create containers of a Distro, so the QGis packages can be isolated from the host system.

GUI app entries are included in the explanations.

---

Then I added an instruction to add missing python dependencies using pip, under the main Linux chapter.

I also added a tutorial how to enforce XWayland, as Wayland shows a warning of missing features. This is opt-in but automated.

PR #1243, opened 2024-05-03
Removed Fedora alterative software repositories because the dani/qgis…

rnanclares:
… copr repository is inactive and hasn't produce an usable build for a year. 

I've removed the copr dani/qgis repository recommendation because it's been inactive for a whole year. Check here: https://copr.fedorainfracloud.org/coprs/dani/qgis/


agiudiceandrea:
See also https://github.com/qgis/QGIS-Website/pull/1196.

PR #1249, opened 2024-05-27
Typo fixes

AlkoholLiga:
Fixes some typos in german translation


DelazJ:
@AlkoholLiga Thanks for the fixes but translation work is done on Transifex platform and then po files are pulled back to this repository. So any fixes to translations should be done there, otherwise they will likely be overwritten at the next pull. Please [join the german team](https://qgis.org/de/site/getinvolved/translate.html#becoming-a-translator).

PR #1254, opened 2024-06-12
Fix bad URL & logo in commercial support

sylvainbeo:


sylvainbeo:
Hi @DelazJ @timlinux , could someone have a look at this fix ?

Thank you

andreasneumann:
While we are at it - could we also switch to https for the Oslandia link?

sylvainbeo:
Indeed. This is fixed

